PR #13, opened 2019-12-11
bug fix

romainVala:


fepegar:
The first commit was already fixed when I edited your previous PR.

I'd rather leave to the user the responsibility of reading the landmarks and just get an `ndarray` as input.

fepegar:
Also, it'd be great if you could open an issue before starting to modify things so that we can discuss whether it's worth it to spend time on it.

PR #39, opened 2020-01-11
Increase masking flexibility for normalization

### LINKED ISSUE #38: Possibility to use external mask when masking function
For some Transform : histogram_standardization rescale z_normalization there may be an interest to use a mask (just to compute percentyl or min max values) then the transform is apply to the all images

This is possible with histogram_standardization (and we should add the option with rescale)

My point is that a masking_function is not an easy job, and I do prefer to use mask compute independently (with your prefered software)

In the case of external mask, you can easily use them by including then in the ImageDataset

```python
suj = [[
    Image('T1','/oneT1.nii.gz',INTENSITY),
    Image('mask','/coresponding_mask.nii.gz',LABEL)
     ]]
```

Then you can use a parameter mask_field_name to get this mask when needed

here is an example of such an implementation 
  https://github.com/romainVala/torchio/blob/master/torchio/transforms/rescale_within_mask.py

but I thing it is better, to directly incorporate it as an optional feature of the Rescale class with something like 

```python
   if sefl.mask_field_name is not None :
       mask_data = sample[self.mask_field_name]['data']
```

About ZNormalization I would prefer to set use_mean_threshold to False by default and add the possiblity to use either a masking_function or the external mask 

For histogram_standardization, a masking_function is expected, (but not use in the call), I added the possibility of external mask 
   https://github.com/romainVala/torchio/blob/master/torchio/transforms/histogram_standardization.py

of course those option are exclusive, and the user should set only one of them, I am not sure what the best way to code it properly, I did something like

```python
    if masking_function is not None:
        mask = masking_function(img)
    else:
        if mask_data is not None:
            mask = mask_data
        else:
            mask = np.ones_like(img, dtype=np.bool)
```

fepegar:
All normalization classes now take a kwarg names `masking_method` that can be:
1. `None`: mask is all ones
2. A string: the sample is queried for the corresponding image name
3. A Python callable such as `torchio.transforms.NormalizationTransform.mean`

Resolves #38.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,11 +72,12 @@ The paths suffix must be `.nii`, `.nii.gz` or `.nrrd`.
 import torchio
 
 subject_a = [
-    Image('t1', '~/Dropbox/MRI/t1.nii.gz', torchio.INTENSITY),
+    Image('t1', '~/Dropbox/MRI/t1.nrrd', torchio.INTENSITY),
     Image('label', '~/Dropbox/MRI/t1_seg.nii.gz', torchio.LABEL),
 ]
 subject_b = [
     Image('t1', '/tmp/colin27_t1_tal_lin.nii.gz', torchio.INTENSITY),
+    Image('t2', '/tmp/colin27_t2_tal_lin.nii', torchio.INTENSITY),
     Image('label', '/tmp/colin27_seg1.nii.gz', torchio.LABEL),
 ]
 subjects_list = [subject_a, subject_b]
--- a/README.md
+++ b/README.md
@@ -162,7 +162,8 @@ A transform can be quickly applied to an image file using the command-line
 tool `torchio-transform`:
 
 ```shell
-$ torchio-transform input.nii.gz RandomMotion output.nii.gz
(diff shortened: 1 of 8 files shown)


fepegar:
@romainVala

romainVala:
Thanks for this quick improvment,
Nice propose solution !

I just wander, if you should make the get_mask more flexible
insetead of :
return  sample[self.mask_name]['data'].bool()

I would put something like
return  sample[self.mask_name]['data'] > self.mask_name_threshold

with default value 0 for  self.mask_name_threshold

The idea, is that in practice we often have mask that are not really binary, this gives the user, more flexibility to define the threshold for binarising the mask, and the 0 default will handle the natural way of considering mask .

fepegar:
I'm not sure. I think the user should be responsible for creating a proper binary mask, with only ones and zeros. If it doesn't have only ones and zeros, it's not really a mask but a probability map, a label map or whatever. I don't think creating the mask beforehand should give any trouble, as a compressed binary mask usually takes only some KB of storage.

PR #92, opened 2020-02-28
Generalize input/output logic

### LINKED ISSUE #82: Generalize io module
Code to write NRRD files is missing:
https://github.com/fepegar/torchio/blob/128895a88cdb1165f8b8533e11835a6b1db12b8c/torchio/data/io.py#L83-L88

fepegar:
Resolves #82.

### DIFF
--- a/torchio/data/io.py
+++ b/torchio/data/io.py
@@ -66,8 +66,8 @@ def _write_nibabel(
         path: TypePath,
         ) -> None:
     """
-    Expects a path with .nii or .nii.gz extension, and writes a NIfTI-1 file
-    with qform only
+    Expects a path with an extension that can be used by nibabel.save
+    to write a NIfTI-1 image, such as '.nii.gz' or '.img'
     """
     nii = nib.Nifti1Image(tensor.numpy(), affine)
     nii.header['qform_code'] = 1
--- a/torchio/data/io.py
+++ b/torchio/data/io.py
@@ -1,54 +1,44 @@
 from pathlib import Path
 from typing import Union, Tuple, TypeVar
-import nrrd
 import torch
 import numpy as np
 import nibabel as nib
+import SimpleITK as sitk
 from .. import TypePath, TypeData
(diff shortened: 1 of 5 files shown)


torchio/data/io.py
```diff
@@ -1,88 +1,84 @@
...
 
-def _write_nifti(
+def _write_nibabel(
         tensor: torch.Tensor,
         affine: TypeData,
         path: TypePath,
         ) -> None:
+    """
+    Expects a path with .nii or .nii.gz extension, and writes a NIfTI-1 file
+    with qform only
```
bcdarwin: Since Nibabel can theoretically write out other file formats (though often conversion isn't implemented), should be "Expects a path with an extension  such as '.nii' or 'nii.gz' to which `nibabel.save` supports writing Nifti1 files" or similar?

fepegar: Well, the issue is that my code is now written to create a nib.Nifti1Image object. Do you know if there is a way to let NiBabel choose the class depending on the extension?

bcdarwin: Yeah, I'm not sure.  You could try to do something with `nib.imageclasses.class_map` but some of the extensions overlap, ".mnc" maps to an ancient MINC1 class, etc ... maybe it's fine to keep as-is for now and just tweak the docstring slightly?

fepegar: Sounds a bit overkill, especially for those formats that are, I believe, mostly quite old. Do you use MINC? If people ask for it, I'll consider working on this. That `class_map` seems like a good resource if ever needed.

fepegar: I see what you mean by "to which nibabel.save supports writing Nifti1 files". One could also use '.hdr' or '.img.gz' (although why on earth?!).

bcdarwin: Yeah, seems SimpleITK is ahead in terms of format support.  I do use MINC but don't really need special support -- I can just convert first or use SITK.

fepegar:
@bcdarwin, thanks for your insight.

PR #98, opened 2020-03-09
Removed a wrong argument 

ibro45:
In the last commit, somehow the argument 'patches' found it's way inside of the get_output_tensor() method


fepegar:
Oops, good catch. I need to improve my tests ASAP. Thank you!

fepegar:
@all-contributors please add @ibro45 for bug

Issue #104, opened 2020-03-24
crop and pad around mask center


romainVala:
Hello

what do you things of the possibility to do the same as CenterCropOrPad, but instead of take the center of the image I would like to take the centre of a label mask (center of non zero element)

Could be very useful to get the same field of view center on the brain, if you have a brain mask.

It could either be an option in CenerCropOrPad, or a new class ... ?

this may be a little bit tricky since there could be a crop for the lower part and a pad for the upper part ...

++

fepegar:
Hi Romain,
Is this something you would like to use? It does seem a bit complex for a transform and more like something that could be done beforehand.

In case we wanted to implement it, I guess it could be something like: `CenterCropOrPad(padding, around=x)`, where x can be the name of the label image or even a point in voxel or world coordinates.

What if the label image has more than one connected component? Only the largest should be used? What should be used, the centroid of the component or the middle of its bounding box?

Once there is a center, I think the best would be creating a new image with the desired bounds (being careful about computing the origin) and resampling the intensity images with an identity transform using the computed image as reference. Would you like to work on this?

romainVala:
Hi
yes I need it to process T1 'raw' images that have different field of view, and different head position.
I agree usually one use to coregister and reslice to a common template, but since I try to learn the motion, keeping the acquisition plan is importante, so I would like to avoid reslicing

In my case I will apply it with brain mask so only one connected component, and then If I understand correctly the centroid is the same a the middle of the bounding box, no ?

The way I imagine is to find min max index of nozero voxel of the mask. this gives me a mask field of view and I take the center to then compute crop and pad

I do not understand your last point, once I have the center, why do I need to reslice ? I want to compute crop or pad in each side of the center separatly

I can try to make a proposition and let you choose to incorporate it or not. Should I make a new class ?

fepegar:
You're right that reslicing/resampling is not needed. I just thought of it as the fastest way of implementing, but I'm probably wrong/lazy. Note, though, that no interpolation would be needed so the operation would be lossless and very fast.

What I meant with preprocessing is cropping all images before training. But I guess you're planning to apply e.g. rotations on the fly and then you want to crop the image around the brain, wherever it ends up. And that's why you want a transform. Is this correct?

Maybe a general `CropOrPad` class can be used to crop or pad around a point, not necessarily the center.

About the centroid vs bounding box. Consider this image:

```
1 1 1
1 0 0
1 0 0
```

[...]

PR #110, opened 2020-04-01
Adding CropOrPad

GReguig:
Hello,

This time, I am making a cleaner PR for my CropOrPad implementation allowing to either use the center of the volume or the center of a given mask. This comes from the issue : #104 

I hope this helps.


fepegar:
This looks much better, thanks :)

I'll make some small changes and tests and merge ASAP.

torchio/transforms/preprocessing/spatial/cropOrPad.py
```diff
@@ -0,0 +1,151 @@
...
+        return padding_params, cropping_params
+
+    def _compute_mask_center_crop_or_pad(self, sample: dict):
+        mask = sample[self.mask_key][DATA].numpy()
+        # Original sample shape (from mask shape)
+        sample_shape = np.squeeze(mask).shape
+        # Calculate the bounding box of the brain
+        xmin, xmax, ymin, ymax, zmin, zmax = self._bbox_mask(np.squeeze(mask))
+        # Coordinates of the center of the brain
+        center_mask = ((xmax - xmin) / 2 + xmin, (ymax - ymin) / 2 + ymin, (zmax - zmin) / 2 + zmin)
```
fepegar: Can this be vectorized with NumPy?

torchio/transforms/preprocessing/spatial/cropOrPad.py
```diff
@@ -0,0 +1,151 @@
...
+        cropping = []
+        for dim, center_dim in enumerate(center_mask):
+            # Compute the coordinates of the target shape taken from the center of the mask
+            begin = center_dim - (self.bounds_parameters[2 * dim] / 2)
+            end = center_dim + (self.bounds_parameters[2 * dim + 1] / 2)
+            # Check if dimension needs padding (before or after)
+            begin_pad = abs(np.minimum(begin, 0))
+            end_pad = np.maximum(end - sample_shape[dim], 0)
+            # Check if cropping needed
+            begin_crop = abs(np.round(abs(np.maximum(begin, 0)))).astype(np.uint)
```
fepegar: Is the outer `abs` needed?

torchio/transforms/preprocessing/spatial/cropOrPad.py
```diff
@@ -0,0 +1,151 @@
...
+            end_crop = abs(np.round(np.minimum(end - sample_shape[dim], 0))).astype(np.uint)
```
fepegar: Can you replace all `np.round` by `round`? That way, you won't need to convert to `int`. Also, I think you can use `min` instead of `np.minimum`.

fepegar: And `max` instead of `np.maximum`.

torchio/transforms/preprocessing/spatial/cropOrPad.py
```diff
@@ -0,0 +1,151 @@
...
+            # Add padding values of the dim to the list
+            padding.append(np.round(begin_pad).astype(np.uint))
+            padding.append(np.round(end_pad).astype(np.uint))
```
fepegar: Same as above. Also, why not rounding above, as for cropping?

torchio/transforms/preprocessing/spatial/cropOrPad.py
GReguig: Which part do you wish to vectorize ? The calculation of the center given a min and max ?

torchio/transforms/preprocessing/spatial/cropOrPad.py
GReguig: You are right, for a maximum with 0, there's no need for any `abs` operation

torchio/transforms/preprocessing/spatial/cropOrPad.py
GReguig: Actually, the returned values from all these operations (without `np`) are floats. I will replace these operations but it will still require a final casting operation

torchio/transforms/preprocessing/spatial/cropOrPad.py
GReguig: Yes, it will be cleaner this way.

torchio/transforms/preprocessing/spatial/crop_or_pad.py
```diff
@@ -102,12 +102,12 @@ def _compute_center_crop_or_pad(self, sample: dict):
...
         cropping = -np.minimum(diff_shape, 0)
-        if cropping.any():
-            cropping_params = self._get_six_bounds_parameters(cropping)
+        #if cropping.any():
+        cropping_params = self._get_six_bounds_parameters(cropping) if cropping.any() else None
 
         padding = np.maximum(diff_shape, 0)
-        if padding.any():
-            padding_params = self._get_six_bounds_parameters(padding)
+        #if padding.any():
```
GReguig: I added a default value `None` if no padding or cropping is needed. This raised up a bug during my testing. There might be a cleaner, more generic way to handle this ?

torchio/transforms/preprocessing/spatial/crop_or_pad.py
```diff
@@ -132,24 +132,32 @@ def _compute_mask_center_crop_or_pad(self, sample: dict):
...
     def apply_transform(self, sample: dict) -> dict:
         padding_params, cropping_params = self.compute_crop_or_pad(sample)
         padding_kwargs = dict(
             padding_mode=self.padding_mode, fill=self.padding_fill)
-        sample = Pad(padding_params, **padding_kwargs)(sample)
-        sample = Crop(cropping_params)(sample)
+        if padding_params is not None:
+            sample = Pad(padding_params, **padding_kwargs)(sample)
+        if cropping_params is not None:
+            sample = Crop(cropping_params)(sample)
```
GReguig, line 156: Just a check whether or not the cropping/padding operation is needed

GReguig:
Hello, thanks for the review. I made a few changes (I think this is a bit cleaner this way) but there is certainly room for improvement.

torchio/transforms/preprocessing/spatial/cropOrPad.py
fepegar: What about something like

```python
# Make self._bbox_mask return two NumPy arrays
bb_min, bb_max = self._bbox_mask(mask[0])  # squeeze might remove other dimensions than the first
center_mask = (bb_max - bb_min) / 2 + bb_min
```

torchio/transforms/preprocessing/spatial/crop_or_pad.py
fepegar: This looks ok to me.

torchio/transforms/preprocessing/spatial/cropOrPad.py
fepegar: Did you end up removing the casting?

GReguig: I did not remove it, I moved further in the function (it does not work otherwise)

torchio/transforms/preprocessing/spatial/cropOrPad.py
GReguig: Ok, it is done.

torchio/transforms/preprocessing/spatial/crop_or_pad.py
```diff
@@ -79,7 +79,7 @@ def _bbox_mask(mask_volume: np.ndarray):
         rmin, rmax = np.where(r)[0][[0, -1]]
         cmin, cmax = np.where(c)[0][[0, -1]]
         zmin, zmax = np.where(z)[0][[0, -1]]
-        return rmin, rmax, cmin, cmax, zmin, zmax
+        return np.array([rmin, cmin, zmin,]), np.array([cmax,rmax, zmax])
```
fepegar, line 88: Can you fix the max coordinates?

GReguig: Sorry about that. It should be OK now.

fepegar:
What happens if the mask is empty?

fepegar:
I'm running the following test:

```python
    def test_center_mask(self):
        """The mask bounding box and the input image have the same center"""
        target_shape = 8, 22, 30
        transform_center = torchio.transforms.CropOrPad(target_shape, padding_fill=2)
        transform_mask = torchio.transforms.CropOrPad(
            target_shape,
            mask_name='label',
            padding_fill=2,
        )
        mask = self.sample['label'][DATA]
        mask *= 0
        mask[0, 4:6, 9:11, 14:16] = 1
        transformed_center = transform_center(self.sample)
        transformed_mask = transform_mask(self.sample)
        zipped = zip(transformed_center.values(), transformed_mask.values())
        for image_center, image_mask in zipped:
            import SimpleITK as sitk
            from torchio.utils import nib_to_sitk
            sitk.WriteImage(nib_to_sitk(self.sample['t1'][DATA], self.sample['t1'][AFFINE]), '/tmp/a_orig.nii')
            sitk.WriteImage(nib_to_sitk(self.sample['label'][DATA], self.sample['label'][AFFINE]), '/tmp/a_label.nii')
            sitk.WriteImage(nib_to_sitk(image_center[DATA], image_center[AFFINE]), '/tmp/a_center.nii')
            sitk.WriteImage(nib_to_sitk(image_mask[DATA], image_mask[AFFINE]), '/tmp/a_mask.nii')
            break
            assert_array_equal(image_center[DATA], image_mask[DATA])
            assert_array_equal(image_center[AFFINE], image_mask[AFFINE])
```

These are the saved images:
Screenshot

As you see, the padding using the mask is not consistent with the center mode. Can you please fix this?

I'll push some tests so that you can run them locally.

fepegar:
I don't know why the tests are not shown here, but here they are: https://travis-ci.org/github/fepegar/torchio

romainVala:
there may be an error in you test :
since the  target_shape = 8, 22, 30 the image center is 2, 11, 15
so to have the same center for the mask you should replace
 mask[0, 4:6, 9:11, 14:16] = 1
by
 mask[0, 3:6, 10:13, 14:17] = 1
no ?

romainVala:
Ah no, target shape is the objectiv,
what is the image shape (before the crop ) ?

fepegar:
```
# Empty mask
0 0 0 0 0 0 0 0 0 0
         ^
# Center at 4.5

# Fill 3:6
0 0 0 1 1 1 0 0 0 0
        ^
# Center at 4

# Fill 4:6
0 0 0 0 1 1 0 0 0 0
         ^
# Center at 4.5
```

romainVala:
ok you're rigth, 
there may be a rounding problem in the code, since you pick an example where the center is exactly between 2 pixel ...

fepegar:
Yes, I thought that it might be related to rounding. Note that rounding can be tricky:
```python
>>> import numpy as np
>>> round(3.5)
4
>>> round(4.5)
4
>>> np.round(3.5)
4.0
>>> np.round(4.5)
4.0
```

romainVala:
that's funny, it prefers even number !

fepegar:
Yes. [This variant of the round-to-nearest method is also called convergent rounding, statistician's rounding, Dutch rounding, Gaussian rounding, odd–even rounding, or bankers' rounding](https://en.wikipedia.org/wiki/Rounding#Round_half_to_even) :D

GReguig:
Hello,

I think I found a little fix for this problem which is to round directly the center coordinates prior to the calculations. However, I used a personal rounding operation in order to avoid the problem you mentioned. I guess it's not the cleanest way but since the native python function does not work as expected, it is the easiest way I found. 
`
def exact_round(value):
    """
    Computes the exact round value of the argument
    Args:
        value: the value to round

    Returns: the rounded value of :attr:`value`

    """
    decimals = value % 1
    if decimals >= .5:
        return int(value//1 + 1)
    else:
        return int(value)
`
Using this function, I modified the function `_compute_mask_center_crop_or_pad` (line 180)

`        for dim, center_dimension in enumerate(center_mask):
            center_dim = exact_round(center_dimension)
`

@fepegar If you agree with this kind of implementation, I will push this version

fepegar:
As long as it works! See another method here: https://stackoverflow.com/a/33019948/3956024

You can run the tests with, for example `pytest`, before pushing. Although we'll need to write more exhaustive tests to make sure it works as expected.

GReguig:
I did run the tests manually (not with `pytest`). Everything seems to work fine. 

Also, I have a suggestion about the case where the key of the mask is not fine. What do you think about raising a warning instead of an error and executing the center mode ? This would allow to treat a dataset with missing masks without raising the error and stopping the execution.

torchio/utils.py
```diff
@@ -206,3 +206,19 @@ def sitk_to_nib(image: sitk.Image) -> Tuple[np.ndarray, np.ndarray]:
     affine[:3, :3] = RZS
     affine[:3, 3] = translation
     return data, affine
+
+
+def exact_round(value):
```
fepegar, line 211: Can you rename this to `round_up` and add typing hints? I.e. `def round_up(value: float) -> float:` and explain that it rounds toward infinity? This method is not really more exact, it's just a different convention to Python's.

fepegar:
> Also, I have a suggestion about the case where the key of the mask is not found. What do you think about raising a warning instead of an error and executing the center mode ? This would allow to treat a dataset with missing masks without raising the error and stopping the execution.

Sounds good. In that case, `test_crop_pad` needs to be modified as well.

torchio/utils.py
GReguig: Done

fepegar:
Please see https://github.com/fepegar/torchio/pull/110#issuecomment-608372579

GReguig:
Sorry, I skipped this one. 

What is an empty mask ? From what I saw in the implementation, it is not possible to just have a `None` value. Do you mean a mask where all values are `0` ?

fepegar:
> Do you mean a mask where all values are `0` ?

Yes.

tests/transforms/preprocessing/test_crop_pad.py
```diff
@@ -59,7 +59,12 @@ def test_deprecation(self):
             CenterCropOrPad(1)
 
     def test_empty_mask(self):
-        pass
+        target_shape = 8, 22, 30
+        transform = CropOrPad(target_shape=target_shape, mode="mask", mask_name="label")
```
fepegar, line 63: This will make the tests fail as I removed the `mode` kwarg.

torchio/transforms/preprocessing/spatial/crop_or_pad.py
```diff
@@ -167,10 +167,19 @@ def _compute_mask_center_crop_or_pad(self, sample: dict):
                 f' not found in sample keys: {tuple(sample.keys())}'
                 f'. Using the center of the volume for cropping/padding instead'
             )
-            warnings.warn(message)
+            warnings.warn(message=message)
```
fepegar, line 170: Why this replacement?

tests/transforms/preprocessing/test_crop_pad.py
fepegar: Actually, I forgot to remove it. I'll modify this.

fepegar:
I think we're good to go.

fepegar:
Thanks for your contribution, Ghiles! I hope it wasn't too painful.

torchio/transforms/preprocessing/spatial/crop_or_pad.py
GReguig: No particular reason, I just tend to use kwargs instead of args. Sorry about that

GReguig:
Cool, I am glad it finally got merged !

PR #114, opened 2020-04-03
Add duecredit items

### LINKED ISSUE #101: zenodo and duecredit
Hi,

Scientific citation is always nice, but there is always a way to improve it. :)
First, you could deposit your code at zenodo.org.
It has a github integration, so all you need to do after connecting the services
is just creating a git release, and it will automatically deposited at Zenodo.
In this way people can refer to specific versions easily.
https://zenodo.org/

In general, Zenodo is very permissive and free of use (EU/CERN project).
You can deposit anything (source code, pdf, any files; by topic: finished papers, drafts, codes, data, etc.), and you can attach external DOIs, if your paper gets published later, you can use multiple versions, etc. 

Related: Duecredit is a lesser known but nice python library for embedding citation
information. The main idea that if you run a code with duecredit, it will collect
stat from all the imported projects, and emitt a list what you should cite.
So as a library author: you can annotate your main library with your citation information,
but you can also refer to implementations. E.g. if in function F use use a specific method,
you can annotate it. 

If your downstream users don't use duecredit, then  nothing happens, so it does not give
an extra dependency users. (Think about it like flake8 or coverage: if someone uses it, it gives information about the code, but it is not a depencency of the project.) 
https://github.com/duecredit/duecredit
But i think in CS we should lead in terms of automated credits, so i recommend duecredit.

fepegar:
Resolves #101.

### DIFF
--- a/torchio/__init__.py
+++ b/torchio/__init__.py
@@ -12,7 +12,6 @@ from . import datasets
 from . import reference
 
 print('If you use TorchIO for your research, please cite the following paper:')
-print('\nPérez-García et al., TorchIO: a Python library for efficient loading,')
+print('Pérez-García et al., TorchIO: a Python library for efficient loading,')
 print('preprocessing, augmentation and patch-based sampling of medical images')
-print('in deep learning')
-print('(https://arxiv.org/abs/2003.04696)')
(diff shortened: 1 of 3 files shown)


fepegar:
@all-contributors please add @dvolgyes for ideas

Issue #112, opened 2020-04-02
Allow non-image type label in subject


nwschurink:
**Is your feature request related to a problem? Please describe.**
I am planning on building a classification network, where the network needs to classify multimodality images of tumor into either benign or malignant. For each tumor I have 3 images e.g. T1, T2 and DWI and each tumor has 1 binary label associated to it e.g. [0,1]. 

At this moment it is only possible to add a 'label' of the Class "Image" to a subject. Which in my case does not make sense as this is not a segmentation but a classification task. 

**Describe the solution you'd like**
Allow the Subject class to also accept a new Label class, which can contain a string, int, list or something like that.

OR

Another easy to implement option would be to allow setting an optional label attribute to the class in the init. Something in the sense of:

```
def __init__(self, *images: Image, name: str = '', label: TypeLabel):
        self._parse_images(images)
        super().__init__(images)
        self.name = name
        self.label = label
```

With ` TypeLabel = Union[str, int, float]` or something similar to that.

**Describe alternatives you've considered**
I have considered creating my own labellist, but it makes more sense to integrate the label into an instance of Subject.

fepegar:
Hi @nwschurink,

I'm planning to add a classification example soon. That will probably give me some ideas to improve the library for the task.

For now, I think you can create a class that inherits from `ImagesDataset`. Would this work for you?

```python
import torchio

class ClassificationDataset(torchio.ImagesDataset):
    def get_sample_dict_from_subject(self, subject: Subject):
        sample = super().get_sample_dict_from_subject(subject)
        sample['diagnosis'] = self.get_label_from_subject()
        return sample

    def get_label_from_subject(self, subject):
        label = get_label(subject.name)  # defined somewhere else


subject = torchio.Subject(
    torchio.Image('t1', 'subject_a.t1.nii.gz', torchio.INTENSITY),
    torchio.Image('t2', 'subject_a.t2.nii.gz', torchio.INTENSITY),
    torchio.Image('dwi', 'subject_a.dwi.nii.gz', torchio.INTENSITY),
    name='subject_a',
)

subjects = [subject]

dataset = ClassificationDataset(subjects)
sample = dataset[0]
label = sample['diagnosis']
```

Maybe I can add `**kwargs` to `Image` and `Subject`, in case the user wants to add custom fields to the subject samples or the images within.

nwschurink:
Hi @fepegar ,

[...]

PR #115, opened 2020-04-03
Added **kwargs to dictionary of Subject and Image

nwschurink:
Edited the `__init__` of `Image` and `Subjec`t such that `**kwargs` get assigned to the dictionary of the class.

I added name again as a positional argument, as the name attribute is used in several subroutines and cannot be undefined.

Reference: #112.

### DIFF
--- a/torchio/data/images.py
+++ b/torchio/data/images.py
@@ -41,7 +41,7 @@ class Image:
         self.name = name
         self.path = self._parse_path(path)
         self.type = type_
-        self.kwargs = kwargs
+        self.__dict__.update(kwargs)
 
     def _parse_path(self, path: TypePath) -> Path:
         try:
@@ -87,10 +87,11 @@ class Subject(list):
         **kwargs: Items that will be added to the subject sample.
     """
 
-    def __init__(self, *images: Image, **kwargs):
+    def __init__(self, *images: Image, name: str = '', **kwargs):
         self._parse_images(images)
         super().__init__(images)
-        self.kwargs = kwargs
+        self.name = name
+        self.__dict__.update(kwargs)
 
     def __repr__(self):
         return f'{__class__.__name__}("{self.name}", {len(self)} images)'
--- a/torchio/data/images.py
+++ b/torchio/data/images.py
@@ -24,8 +24,6 @@ class Image:
     r"""Class to store information about an image.
 
     Args:
-        name: String corresponding to the name of the image, e.g. ``t1``,
-            or ``segmentation``.
         path: Path to a file that can be read by
             :mod:`SimpleITK` or :mod:`nibabel` or to a directory containing
             DICOM files.
@@ -37,8 +35,7 @@ class Image:
             subject sample.
     """
 
-    def __init__(self, name: str, path: TypePath, type_: str, **kwargs):
-        self.name = name
+    def __init__(self, path: TypePath, type_: str, **kwargs):
         self.path = self._parse_path(path)
         self.type = type_
         self.__dict__.update(kwargs)
@@ -50,11 +47,7 @@ class Image:
             message = f'Conversion to path not possible for variable: {path}'
             raise TypeError(message)
         if not (path.is_file() or path.is_dir()):  # might be a dir with DICOM
-            message = (
-                f'File for image "{self.name}"'
-                f' not found: "{path}"'
-                )
-            raise FileNotFoundError(message)
+            raise FileNotFoundError(f'File not found: {path}')
         return path
 
     def load(self, check_nans: bool = True) -> Tuple[torch.Tensor, np.ndarray]:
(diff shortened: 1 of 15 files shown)


torchio/data/images.py
```diff
@@ -87,10 +87,11 @@ class Subject(list):
         **kwargs: Items that will be added to the subject sample.
     """
 
-    def __init__(self, *images: Image, **kwargs):
+    def __init__(self, *images: Image, name: str = '', **kwargs):
```
fepegar, line 90: Can you explain why you're keeping `name`?

torchio/data/images.py
```diff
@@ -41,7 +41,7 @@ def __init__(self, name: str, path: TypePath, type_: str, **kwargs):
         self.name = name
         self.path = self._parse_path(path)
         self.type = type_
-        self.kwargs = kwargs
```
fepegar, line 44: This attribute is referenced in `ImagesDataset` so I guess it should stay around.

fepegar: Also, if there `kwargs` has many items, users would need many lines of code to add the parameters to the sample. This way, it's just a single line.

torchio/data/images.py
fepegar: I guess you mean it's used in `Subject.__repr__`. That's not very important, so we can just get rid of it.

torchio/data/images.py
nwschurink: I missed the reference in ImageDataset, but what do you mean with that your need many likes of code to add parameters to a sample? Do you mean, like when you'd like a list of labels from an ImageDataset instance?

nwschurink: And what if Subject inherits from 'dict' instead of 'list'? Then you could easily iterate though Subject.keys()

fepegar: I'm not sure that would be necessary. Why not just keeping `kwargs`?

nwschurink: Hmm yea you are right it's not really necessary, maybe it's just a matter of taste then. Accessing Subject.kwargs['key'] feels a bit awkward as I would say the keyword arguments are attributes of the Subject. However, it works perfectly fine this way so it is a good sollution ;) I'll just work with it like this

fepegar: Maybe there's a misunderstanding. I propose keeping the kwargs attribute **and** doing the `Subject.__dict__.update(kwargs)`. So you can still do `subject.diagnose` or, if needed, `subject.kwargs['diagnose']`.

If `Subject` inherits from `dict`, I guess then there's no need for `Image` to have a `name`. Maybe that makes more sense. But what would be the arguments of `Subject`? A dictionary similar to the `sample_dict` would make sense, although it might be a bit more verbose than the current approach.

fepegar:
```python
subject = Subject(
    t1=Image(path, type),
    t2=Image(path, type),
    age=25,
)

sample = {}
for key, value in subject.items():
    if isinstance(value, Image):
        sample[key] = get_image_dict_from_image(value)
    else:
        sample[key] = value
```

Maybe a similar approach could be useful for `Image` as well. But I'm afraid it will be hard to collate the batch by the `DataLoader` if keys start to differ between samples.

@romainVala, what do you think?

romainVala:
hi there, 
yes this thread is related to different point we discussed previously, so I try to follow ..
Until recently (https://github.com/fepegar/torchio/issues/116#issue-594892916) I  was not using Subject, so I am not sure, why to we need such a level.

I just resume to be sure I correctly understand
So your last proposition is changing the definition, but not the end result : dataset is still producing sample dictionary where the keys are the different image type. rigth ?

given that, it is more clear to keep the image name (which will be the key to get the image from the sample to the image level)

Now about adding extra parameters at the subject level, I wonder how you will use them, within a dataloader. everything is rearrange with batch size, so you can not match subject extra parameter ... no?
I thing you can not avoid the collate function, done by pytorch when creating the batch, since this is mandatory to have you extra parameter, still in correspondence with your batch size

does it make sense ?

romainVala:
Although I do understand it make sense to have extra parameter at the subject level, since they may applied to all images, (for instance patient pathology, it not related to T1 and T2, but to both).

In your proposition the subject extra info is added to the sample, so it is fine (ie it will be collate in the batch). that's good

I am just not sure, why to remove the 'T1' key from the Image ...

fepegar:
> I was not using Subject, so I am not sure why we need such a level

Mostly for classification/regression tasks of single numbers, I suppose. 

> So your last proposition is changing the definition, but not the end result : dataset is still producing sample dictionary where the keys are the different image type. right ?

Results should be the same, yes. The keys would be the image *name*.

> I am just not sure, why to remove the 'T1' key from the Image ...

If it's in the Subject, it doesn't need to be in the Image. I think this makes also more sense conceptually, as the image itself doesn't need to know its own name. Also it's consistent with how the sample dictionary is created. Subject and Image are just interfaces to create the dictionaries, so it seems logical that they somehow share part of the key-value structure.

torchio/data/images.py
nwschurink: Having both options of `subject.kwargs['diagnose']` and `subject.diagnose` would be a solution. It would also be really easy to add.

As I'm reading [online](https://stackoverflow.com/questions/2033150/subclassing-dict-should-dict-init-be-called), I am starting to doubt a little bit on whether it makes sense to inherrit from `dict`. Apparently if you'd do, then it would be good practice to make the `Subject.__init__` function similar to how `dict.__init__` works. I think that would mean that the current initialization should be altered such that `Subject.__init__` should accept `key:value` pairs instead of positional arguments. Which would mean that it break all code written with previous versions right?

fepegar: Yes. That's what I propose in my [comment below](https://github.com/fepegar/torchio/pull/115#issuecomment-609696750). Let's continue the discussion there, please.

Don't worry about breaking the code: "[Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable](https://semver.org/#spec-item-4)".

If it makes the library better, I think some disruption is acceptable in this experimental phase.

fepegar:
(See also discussion here: https://github.com/fepegar/torchio/pull/115#discussion_r404002739)

romainVala:
I better understand, (sorry for my previous confusion)
Yes it make sense then,

romainVala:
> ```python
> sample = {}
> for key, value in subject.items():
>     if isinstance(value, Image):
>         sample[key] = get_image_dict_from_image(value)
>     else:
>         sample[key] = value
> ```

where will be this part of the code  ?

fepegar:
> where will be this part of the code ?

In this method:
https://github.com/fepegar/torchio/blob/22459930364374dd8be9d11ab5a0894722c7bb0d/torchio/data/images.py#L210-L221

nwschurink:
Cool, this would indeed fix it! Should I add it to the PR? or do you edit it yourself?

fepegar:
I'll give it a try.

fepegar:
@romainVala @nwschurink can you take a look?

romainVala:
Hi
great job, quite a lot of change, indeed, but always similar, so I am ready to do the same on my script.
seems good to me

nwschurink:
Good morning,

Looks very good, onlys one thing that i'm wondering about. In `Subject.__repr__` a reference is made to `Subject.name` https://github.com/fepegar/torchio/pull/115/files/a92f075848fd7216704db4672699dbe43a649b33..3ceb0fcd191af252aac5ca934fca0653a4dc6095#diff-87328b3066b813d9ca439505a520b7b3R108, however, if `Subject.name` is undefined this will throw an exception.

I'm not sure if you'd want to enforce a name being set for each subject. If so something in the line of this could be added to the beginning of init just before this line https://github.com/fepegar/torchio/pull/115/files/a92f075848fd7216704db4672699dbe43a649b33..3ceb0fcd191af252aac5ca934fca0653a4dc6095#diff-87328b3066b813d9ca439505a520b7b3R99:

```
try:
    self.name = kwargs.pop('name')
except KeyError:
    raise KeyError('a "name" keyword argument must be supplied')
super().__init__(**kwargs)
```

Or if you'd don't want name to be a required argument you'd probably want to define it as an empty string (or delete the reference in `__repr__`) before you call `super().__init__`:

```
self.name = ''
super().__init__(**kwargs)
```

That way if `name` is defined as a `kwarg` it will overwrite `self.name`, and will be defined as an empty string otherwise.

nwschurink:
Ah, never mind, my comment doesn't make sense.. `__name__` refers to the name of the module not the subject name. So the line will work ;) Learned something new.

Anyway, I'd say the code is good to go then

Issue #118, opened 2020-04-08
Add reference image/grid option to resample function


nwschurink:
**Is your feature request related to a problem? Please describe.**
In my dataset I have images with different voxelgrids/spacings. As a result if e.g. a `mask` and `T2` image get resampled to 1x1x1mm voxels, it is possible that the physical locations of these voxel do not match. This happens for example when a mask is made on a different MRI sequence with a lower resolution, such as a `DWI` scan.

It would be nice if there is an option to resample all images to 1 specific reference image. (e.g. the `T2` voxelgrid)

**Describe the solution you'd like**
When resampling using ITK you can set a reference image, such that the voxelgrids overlap.

```
resampler = sitk.ResampleImageFilter()
resampler.SetInterpolator(sitk.sitkLinear)
resampler.SetReferenceImage(reference_image)
mask_image_resampled_to_reference_spacing = resampler.Execute(mask_image)
```

**Describe alternatives you've considered**
Pre-processing all my data outside torchio, however this would probably be a nice feature to add.

**Additional context**
I could take a look at how to implement this into the existing code. I'm not sure whether you would want such an option in the  `torchio.transforms.Resample` function (e.g. with something like a `reference_image='T2'` argument), or if this would be better implemented in a seperate transform.

romainVala:
I like the proposition to add a target space for the resample (I do it with nibabel, with resample_from_to) but I do not understand why there is a mismatch
I can imagine this mismatch happen since you have different field of view
using a composition of resampled to 1mm and crop to get the same field of view does not solve it ?

Never the less I thing it is a good idea to allow a reference space

fepegar:
I think what @nwschurink has two advantages:
1. Transform is faster
1. Information is not lost at interpolation

This can be applied to a sample containing e.g. a T1 and a T2 that are aligned but not in the same grid. The transform can be `Resample(target='T1')`, and the resampling strategy will be chosen according to the type of `target`.

nwschurink:
Yes, this happens when the T1 and T2 are aligned but not in the same grid. The patient coordinatesystem is the same for both scans, however the voxelgrids are different. 

There are many cases when this could happen, e.g. different voxelsizes, slice direction, field of view etc. 

[...]

PR #121, opened 2020-04-09
118 resample to target spacing

nwschurink:
In the resampling transform, I have added the possibility to resample all images to a common voxelgrid as discussed in issue #118  

It's the first time I have been working directly on a package, so if you have any comments on the code I'm glad to get some feedback ;)

### DIFF
--- a/tests/transforms/preprocessing/test_resample.py
+++ b/tests/transforms/preprocessing/test_resample.py
@@ -1,0 +1,45 @@
+from numpy.testing import assert_array_equal
+from torchio import DATA, AFFINE
+from torchio.transforms import Resample
+from torchio.utils import nib_to_sitk
+from ...utils import TorchioTestCase
+
+
+class TestResample(TorchioTestCase):
+    """Tests for `Resample`."""
+    def test_spacing(self):
+        # Should this raise an error if sizes are different?
+        spacing = 2
+        transform = Resample(spacing)
+        transformed = transform(self.sample)
+        for image_dict in transformed.values():
+            image = nib_to_sitk(image_dict[DATA], image_dict[AFFINE])
+            self.assertEqual(image.GetSpacing(), 3 * (spacing,))
+
+    def test_reference(self):
+        sample = self.get_inconsistent_sample()
+        reference_name = 't1'
+        transform = Resample(reference_name)
+        transformed = transform(sample)
+        ref_image_dict = sample[reference_name]
+        for image_dict in transformed.values():
+            self.assertEqual(
+                ref_image_dict[DATA].shape, image_dict[DATA].shape)
+            assert_array_equal(ref_image_dict[AFFINE], image_dict[AFFINE])
+
+    def test_wrong_spacing_length(self):
+        with self.assertRaises(ValueError):
+            Resample((1, 2))
+
+    def test_wrong_spacing_value(self):
+        with self.assertRaises(ValueError):
+            Resample(0)
+
+    def test_wrong_target_type(self):
+        with self.assertRaises(ValueError):
+            Resample(None)
+
+    def test_missing_reference(self):
+        transform = Resample('missing')
+        with self.assertRaises(ValueError):
+            transform(self.sample)
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -46,8 +46,16 @@ class TorchioTestCase(unittest.TestCase):
         """Return a sample containing images of different shape."""
         subject = Subject(
             t1=Image(self.get_image_path('t1_d'), INTENSITY),
-            t2=Image(self.get_image_path('t2_d', shape=(10, 20, 31)), INTENSITY),
-            label=Image(self.get_image_path('label_d', binary=True), LABEL),
+            t2=Image(
+                self.get_image_path('t2_d', shape=(10, 20, 31)), INTENSITY),
+            label=Image(
+                self.get_image_path(
+                    'label_d',
+                    shape=(8, 17, 25),
+                    binary=True,
+                ),
+                LABEL,
+            ),
         )
         subjects_list = [subject]
         dataset = ImagesDataset(subjects_list)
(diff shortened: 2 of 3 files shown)


nwschurink:
Ah, I just see that I forgot to add one check.

When target_spacing=None and a reference_spacing is given, it will check the input and resample everything to the reference spacing. However, when both target_spacing and reference_spacing are None it will try to perform regular resampling with an undefined target_spacing.

Can I edit this using the github code review?

fepegar:
Could you please squash into a single commit and force-push? That way I can see the changes more easily. Here's some help: https://github.com/Slicer/Slicer/pull/4789#issuecomment-606791186

fepegar:
Looking good! I'm going to make some changes. We should have some tests, too.

fepegar:
I refactored a bit and added some tests.

nwschurink:
Code looks good 👍 

I think the warning message would be good. If the images are of different size, you'd run into a lot of issues either with other transforms or in your deep learning pipeline.

So maybe, if a target space is passed and images have a different size the warning could also include a hint that you should first resample to a common voxelgrid right? E.g. if you have overlapping images with different sizes and different voxelgrids you'd need 2 transforms e.g.:

```
transforms =[ torchio.transforms.Resample(target='t1'),
                    torchio.transforms.Resample(target=(1,1,1))]
```

romainVala:
Sorry if I miss the point, but I do not understand, 
if input images have different size, after resample to target, they all will have the target shape, (this is one interest of this transform)
what do I miss ?

romainVala:
ok, I see, the title, retarget='t1' mean you chose the pixel size from the t1 image ? is that correct ?

Why not to make  a "full" resampling ... ie: the resampled image will have the exact same affine as the reference image (this is what nibabel.resample_from_to() function does) ... ?

nwschurink:
@romainVala you are right that if you resample to a target they will have the same target shape. But what if your target doesn't have your desired voxelsize? For my usecase I have images of different sizes that I all want to have in the same field of view (e.g. my `target='t2'`). My target image however has voxelsize `(0.7,0.7,2)` and I want this to be (1,1,1). So that means you'd need to chain the Resample transform.

I understood @fepegar was referring to the case where you have input images of different size, you resample to a certain spacing (e.g. (1,1,1) isotropic) but forget to resampling to a common target first. Then the endresult will be images with 1mm isotropic voxels but will still be of different size?

nwschurink:
I'm not sure if I follow exactly what you're saying. The target='t1' already does a "full" resampling to the 't1' voxelgrid using `nibabel.resample_from_to()` right?

romainVala:
ok thanks for the clarification, we do agree, 
I just though warning was intended, when one use target='t1', but as you said, this is for the case where one use resample to voxel size (1,1,1)
make sense

I can think of one example where the message does not apply
For instance if you do the resample to 1,1,1 and then use CropAndPad (to a specific size), in this case it solve the problem of having different size, (without having to first resample to a target space)

fepegar:
> I can think of one example where the message does not apply
For instance if you do the resample to 1,1,1 and then use CropAndPad (to a specific size), in this case it solve the problem of having different size, (without having to first resample to a target space)

I agree. What about a warning in the docs instead?

nwschurink:
I think the docs would also be good, it's probably the first place people will look :).

nwschurink:
There is still one thing that I'm wondering about. CropAndPad won't accept images of different sizes and will only work when the voxelgrids are already align right?

Because, if you have images with different voxelgrids. If you would resample them to isotropic voxels without also resampling to a common target space and then use CropAndPad. The resampled images will indeed have the same voxelsize and have the same shape after CropAndPad. However a voxel at location (x,y,z) in one image will then not be at the same physical location as voxel (x,y,z) in the other cropped images, because the voxelgrids were defined differently. So in that case you would also need to resample to a common target space before applying CropAndPad right?

romainVala:
The way I would like to use this, is for the case of resampling to a template, 
In this use case, there is no interest to add in each template to the data sample. 
So is it possible to have to possibility to specify a template image as argument ? 
or alternatively, just an an affine (no need to load the data, when only the affine matter !)

the next step is to handel also the case, where you have a "coregistration file" (affine transformation from fsl nifti_reg, ). an other possible argument ?

romainVala:
> There is still one thing that I'm wondering about. CropAndPad won't accept images of different sizes and will only work when the voxelgrids are already align right?

I did not check, but it should work, why not ?

romainVala:
> Because, if you have images with different voxelgrids. If you would resample them to isotropic voxels without also resampling to a common target space and then use CropAndPad. The resampled images will indeed have the same voxelsize and have the same shape after CropAndPad. However a voxel at location (x,y,z) in one image will then not be at the same physical location as voxel (x,y,z) in the other cropped images, because the voxelgrids were defined differently. So in that case you would also need to resample to a common target space before applying CropAndPad right?

it make it possible, but not sure, it will keep the coregistration correctly. ... it depends on the exact preprocessing and data origin
if the matter is to conserve the right coregistration then saffer way is indeed to resample
And if the matter is to have all subject align in the same space, then the extra option I just propose are need

nwschurink:
> So is it possible to have to possibility to specify a template image as argument ?
> or alternatively, just an an affine (no need to load the data, when only the affine matter !)
> 
> the next step is to handel also the case, where you have a "coregistration file" (affine transformation from fsl nifti_reg, ). an other possible argument ?

I agree, this would be a nice additional feature indeed. I can imaging that for your usecase, you would now have to load in this template for every image, having such an argument would probably speed up the transforms.

romainVala:
@fepegar 
you did not comment about the propose extension : 
1_ take a template a argument (either file path, to be load by nibabel, or directly an affine)
2_ take a "coregistration file" as argument, format as an affine, wich need to be multiplied the nifti affine, before to do the resampling

Doing so will allows to have all subjects in the same space,
If you are interested, I can try to make a new PR

fepegar:
I think that maybe a universal `Resample` class is needed that merges `Resample`, `RandomAffine`, `RandomElasticDeformation` and adds the functionality you're proposing, but this should be discussed in a new issue.

romainVala:
not sure : this is quite the opposite use :  here we are in preprocessing, and the randomAffine is used for augmentation

may be the coregistration file is a little out of scope, but very convenient though ...

fepegar:
@all-contributors please add @romainVala for ideas, review

fepegar:
@all-contributors please add @nwschurink for code

Issue #131, opened 2020-04-23
Option to add a coregistration matrix before resampling


GFabien:
**Is your feature request related to a problem? Please describe.**
When working with MRI volumes, we like to have one reference template image and use all the images in the MNI reference space. Therefore, before resampling one image, we need to multiply its affine by a coregistration matrix in order to end up in the MNI reference space. This could be seen as a resampling strategy and therefore we think that it would make sense to add an option in the Resample transform to handle this.

**Describe alternatives you've considered**
To implement this feature we want to add a string parameter to the Resample transform that would point to the name of a key in the images of the dataset. The value associated to this key would be a 4x4 matrix that we would multiply with the affine of the image before resampling. Default matrix would be identity to preserve current behaviour if we do not want to perform coregistration or if there is an image which do not have a coregistration matrix attached.

Usage would be:
```python
import numpy as np
import torchio
from torchio.transforms import Resample
from torchvision.transforms import Compose

coregistration = np.eye(4)
subject = torchio.Subject(
    t1=torchio.Image(img_path, torchio.INTENSITY, coregistration=coregistration),
    label=torchio.Image(mask_path, torchio.LABEL),
    ref=torchio.Image(ref_path, torchio.INTENSITY),
)
subject_list = [subject]
transforms = [
    Resample(target='ref', coregistration='coregistration'),
]
transform = Compose(transforms)
dataset = torchio.ImagesDataset(subject_list, transform=transform)
```

Are you interested in such a feature? I have the PR ready, we implemented this on @romainVala 's fork.

fepegar:
Sure, please go on.

PR #133, opened 2020-04-23
Resample with coregistration (Issue #131)

GFabien:
Implement feature presented in https://github.com/fepegar/torchio/issues/131 by adding a `coregistration` parameter to the Resample transform. This parameter is a string pointing to an attribute of the images the transform is applied to. If this attribute is present, it must be a 4x4 matrix.

Usage:
```python
import numpy as np
import torchio
from torchio.transforms import Resample
from torchvision.transforms import Compose

coregistration = np.eye(4)
subject = torchio.Subject(
    t1=torchio.Image(img_path, torchio.INTENSITY, coregistration=coregistration),
    label=torchio.Image(mask_path, torchio.LABEL),
    ref=torchio.Image(ref_path, torchio.INTENSITY),
)
subject_list = [subject]
transforms = [
    Resample(target='ref', coregistration='coregistration'),
]
dataset = torchio.ImagesDataset(subject_list, transform=transform)
```

### DIFF
--- a/tests/transforms/preprocessing/test_resample.py
+++ b/tests/transforms/preprocessing/test_resample.py
@@ -28,18 +28,23 @@ class TestResample(TorchioTestCase):
                 ref_image_dict[DATA].shape, image_dict[DATA].shape)
             assert_array_equal(ref_image_dict[AFFINE], image_dict[AFFINE])
 
-    def test_coregistration(self):
+    def test_affine(self):
         spacing = 1
-        coregistration_name = 'coregistration'
-        transform = Resample(spacing, coregistration=coregistration_name)
+        affine_name = 'pre_affine'
+        transform = Resample(spacing, pre_affine_name=affine_name)
         transformed = transform(self.sample)
         for image_dict in transformed.values():
-            if coregistration_name in image_dict.keys():
+            if affine_name in image_dict.keys():
                 new_affine = np.eye(4)
                 new_affine[0, 3] = 10
                 assert_array_equal(image_dict[AFFINE], new_affine)
             else:
                 assert_array_equal(image_dict[AFFINE], np.eye(4))
+
+    def test_missing_affine(self):
+        transform = Resample(1, pre_affine_name='missing')
+        with self.assertRaises(ValueError):
+            transform(self.sample)
 
     def test_wrong_spacing_length(self):
         with self.assertRaises(ValueError):
@@ -55,10 +60,5 @@ class TestResample(TorchioTestCase):
 
     def test_missing_reference(self):
         transform = Resample('missing')
-        with self.assertRaises(ValueError):
-            transform(self.sample)
-
-    def test_missing_coregistration(self):
-        transform = Resample(1, coregistration='missing')
         with self.assertRaises(ValueError):
             transform(self.sample)
--- a/tests/transforms/preprocessing/test_resample.py
+++ b/tests/transforms/preprocessing/test_resample.py
@@ -57,3 +57,8 @@ class TestResample(TorchioTestCase):
         transform = Resample('missing')
         with self.assertRaises(ValueError):
             transform(self.sample)
+
+    def test_missing_coregistration(self):
+        transform = Resample(1, coregistration='missing')
+        with self.assertRaises(ValueError):
+            transform(self.sample)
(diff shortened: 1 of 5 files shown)


GFabien:
Sorry, I forgot to remove that.

tests/transforms/preprocessing/test_resample.py
```diff
@@ -27,6 +28,19 @@ def test_reference(self):
...
+        coregistration_name = 'coregistration'
+        transform = Resample(spacing, coregistration=coregistration_name)
+        transformed = transform(self.sample)
+        for image_dict in transformed.values():
+            if coregistration_name in image_dict.keys():
+                new_affine = np.eye(4)
+                new_affine[0, 3] = 10
+                assert_array_equal(image_dict[AFFINE], new_affine)
+            else:
+                assert_array_equal(image_dict[AFFINE], np.eye(4))
```
fepegar, line 42: The transform does not seem to be working, and I think this test is not useful.

The images never contain `coregistration_name`, that goes into the sample dict, so the test jumps directly to line 42, which is True just because the test sample has an identity affine by default.

fepegar: This is the script I used to try the transform:

```python
import numpy as np
from torchio.transforms import Resample
from torchio import Image, ImagesDataset, INTENSITY, Subject

matrix = np.fromstring(
    """1.03324 -0.00579961 -0.0387759 -4.59106
0.00273174 0.939307 0.297793 -5.97419
0.0310681 -0.316461 1.02389 -0.777692
0 0 0 1
""", sep=' ')

subject = Subject(
    t1=Image('~/Dropbox/MRI/t1.nii.gz', INTENSITY),
    mni=Image('~/episurg/mri/mni_icbm152_nl_VI_nifti/icbm_avg_152_t1_tal_nlin_symmetric_VI.nii.gz', INTENSITY),
    to_mni=matrix,
)
subjects_list = [subject]
transform = Resample('mni', coregistration='to_mni')
dataset = ImagesDataset(subjects_list, transform=transform)
transformed = dataset[0]
dataset.save_sample(transformed, dict(t1='/tmp/t1_on_mni.nii.gz'))
```

GFabien: Ok, I will investigate but I can see the effect of the coregistration on the samples I'm looking at, the first is resampled using coregistration while the other one is resampled without it:
with_coregistration
without_coregistration

GFabien: And I have runned the test with a print inside to check and it goes inside the if statement. As I have changed `self.sample`, the t1 image has a coregistration attribute but not the t2.

GFabien: Oh yes, I got the error with your snippet, the coregistration key (in your example "to_mni") must be in the image, not in the subject. It is done this way because it is expected that if a subject has several images they could all have their own coregistration.

fepegar: Oh, you pass the matrix to each image. I thought it was per sample, as you can see in my example. The transform should issue a warning or error if this happens (the argument is not `None`, but the key is not found in any image in the sample). Can you please add that check?

GFabien: Yes you right, I add this check to prevent further problems!

GFabien:
I added the check but I'm not fully convinced that it was the best way to implement it. Tell me if you see a better way to do it.

fepegar:
I made some changes, can you check if you're happy with them?

torchio/transforms/preprocessing/spatial/resample.py
```diff
@@ -94,18 +112,46 @@ def parse_interpolation(interpolation: Interpolation) -> int:
...
 
     def apply_transform(self, sample: Subject) -> dict:
         use_reference = self.reference_image is not None
-        iterable = sample.get_images_dict(intensity_only=False).items()
-        for image_name, image_dict in iterable:
+        use_pre_affine = self.affine_name is not None
+        sample_images = sample.get_images_dict(intensity_only=False)
+        if use_pre_affine:
+            self.check_affine_key_presence(self.affine_name, sample)
+        for image_name, image_dict in sample_images.items():
```
GFabien, line 154: That's a tiny detail but maybe this would be a bit clearer:
```python
use_pre_affine = self.affine_name is not None
if use_pre_affine:
    self.check_affine_key_presence(self.affine_name, sample)
iterable = sample.get_images_dict(intensity_only=False).items()
for image_name, image_dict in iterable:
```

fepegar:
Many thanks for your contribution, @GFabien!

GFabien:
You're welcome, it's a really nice package!

Issue #132, opened 2020-04-23
Resampling using a common reference image for all subjects


GFabien:
**Is your feature request related to a problem? Please describe.**
Currently, resampling to match a reference image is done in the perspective of comparing two modalities of the same subject. Our use case is to compare all subjects to a reference template. Therefore it seems highly redundant to insert the reference image in all subjects and would make more sense to have it directly in the Resample transform.

**Describe the solution you'd like**
We would like that the Resample transform could take care of handling a reference image without expecting it to be an image from the subject.

**Describe alternatives you've considered**
The best solution seems to modify the `target` argument from the Resample transform to make it handle this new use case. We would like it to accept a path but it would conflict with the case of passing a key to the transform as both would be strings and I'm afraid that any strategy to differentiate between keys and paths would be error prone. 

Another solution is to make it accept a tuple that would be `(affine, shape)` as they are the parameters needed by nibabel to perform the resampling. The problem of this solution is that it requires the user to handle the extraction of the affine and the shape from the reference image. 

Do you think adding such option to torchio makes sense? If so do you have suggestions on how to handle it? I can implement it if you think it's relevant.

romainVala:
Personnaly I would like to have both solution implemented : a filepath or (affine,shape)
filepath is the 'lazzy' solution but (affine, shape) has the advantage that you are free to choose the exact shape (independtly of the reference volume, and then avoid an extra CropOrPad transform) 

May be a solution would be to have different inputs arguments name : 

target_key_name
target_filepath
target_affine_shape

and precise in the doc that there are exclusif, only one should be used ...

fepegar:
We could do this:

```
subject = Subject(
    t1=Image('path_to_t1.nii.gz', torchio.INTENSITY),
    ref_path='mni_lin_152.nii.gz',
)
transform = Resample(target='ref_path')
```

And in Resample:
```python
target_value = sample[target_key]
if isinstance(target_value, (str, Path)):
    load image and use as reference
else:
    proceed as usual
```

`CropOrPad` is very cheap, so I don't think we should bother with the `(affine, shape)` option.

PR #134, opened 2020-04-24
Add Path type option for Resample target argument (Issue #132)

GFabien:
Change Resample transform to accept an external reference image as a Path object. Convenient when all samples have to be resampled to a given template. (see https://github.com/fepegar/torchio/issues/132)

Usage:
```python
from pathlib import Path
import torchio
from torchio.transforms import Resample

transform = Resample(Path('template.nii.gz'))
subject = torchio.Subject(t1=torchio.Image('t1.nii.gz', torchio.INTENSITY))
dataset = torchio.ImagesDataset([subject], transform=transform)
```

### DIFF
--- a/tests/transforms/preprocessing/test_resample.py
+++ b/tests/transforms/preprocessing/test_resample.py
@@ -17,7 +17,7 @@ class TestResample(TorchioTestCase):
             image = nib_to_sitk(image_dict[DATA], image_dict[AFFINE])
             self.assertEqual(image.GetSpacing(), 3 * (spacing,))
 
-    def test_reference(self):
+    def test_reference_name(self):
         sample = self.get_inconsistent_sample()
         reference_name = 't1'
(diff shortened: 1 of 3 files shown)


fepegar:
Hi again, @GFabien. I think it's not a good idea to make a decision based on whether the argument is of type `str` or `Path`, as paths are still most often given as strings nowadays.

I think the way to go is passing a key name to the transform (as I suggested [here](https://github.com/fepegar/torchio/issues/132#issuecomment-618316623)). If the value for that key is an `Image`, that is the reference. If it's a `str`/`Path`, that image will be loaded and used as reference.

What do you think?

GFabien:
Ok I think I misunderstood your point. I totally agree with you, here is my new suggestion.

fepegar:
Thakns again @GFabien!

fepegar:
@all-contributors please add @GFabien for code

PR #167, opened 2020-05-26
Make sampler copy non image fields

GFabien:
Here is a small PR to avoir loss of meta data information when creating patches. 

Previous behaviour: only images (of type `INTENSITY` or `LABEL`) were processed and saved in the `cropped_sample` leading to the loss of any non image property of the sample.

New behaviour: all properties are copied and images are updated.


fepegar:
Thanks @GFabien!

PR #173, opened 2020-05-29
Fix random affine not rotating around center

### LINKED ISSUE #172: Random affine is not applied around image center
**🐛Bug**

As reported by @romainVala in #171.

fepegar:
Additionally, I added support to choose whether the rotation is applied
around the origin (in world coordinates) or the image center.

Fixes #172.

### DIFF
--- a/torchio/data/image.py
+++ b/torchio/data/image.py
@@ -4,8 +4,20 @@ from typing import Any, Dict, Tuple, Optional
 
 import torch
 import numpy as np
+import SimpleITK as sitk
 
-from ..torchio import TypePath, DATA, TYPE, AFFINE, PATH, STEM, INTENSITY
+from ..utils import nib_to_sitk
+from ..torchio import (
+    TypePath,
+    TypeTripletInt,
+    TypeTripletFloat,
+    DATA,
+    TYPE,
+    AFFINE,
+    PATH,
+    STEM,
+    INTENSITY,
+)
 from .io import read_image
 
 
@@ -58,18 +70,12 @@ class Image(dict):
         self.is_sample = False  # set to True by ImagesDataset
 
     @property
-    def shape(self):
+    def shape(self) -> Tuple[int, int, int, int]:
         return self[DATA].shape
 
     @property
-    def spatial_shape(self):
+    def spatial_shape(self) -> TypeTripletInt:
         return self.shape[1:]
-
-    def is_2d(self):
-        return self.shape[-3] == 1
-
-    def numpy(self):
-        return self[DATA].numpy()
 
     @staticmethod
     def _parse_path(path: TypePath) -> Path:
@@ -132,3 +138,23 @@ class Image(dict):
         if check_nans and torch.isnan(tensor).any():
             warnings.warn(f'NaNs found in file "{self.path}"')
         return tensor, affine
+
+    def is_2d(self) -> bool:
+        return self.shape[-3] == 1
+
+    def numpy(self) -> np.ndarray:
+        return self[DATA].numpy()
+
+    def as_sitk(self) -> sitk.Image:
+        return nib_to_sitk(self.data, self.affine)
+
+    def get_center(self, lps: bool = False) -> TypeTripletFloat:
+        """Get image center in RAS (default) or LPS coordinates."""
+        image = self.as_sitk()
+        size = np.array(image.GetSize())
+        center_index = (size - 1) / 2
+        l, p, s = image.TransformContinuousIndexToPhysicalPoint(center_index)
+        if lps:
+            return (l, p, s)
+        else:
+            return (-l, -p, s)
--- a/torchio/transforms/augmentation/spatial/random_affine.py
+++ b/torchio/transforms/augmentation/spatial/random_affine.py
@@ -4,7 +4,14 @@ import torch
 import numpy as np
 import SimpleITK as sitk
 from ....data.subject import Subject
-from ....torchio import LABEL, DATA, AFFINE, TYPE, TypeRangeFloat
+from ....torchio import (
+    LABEL,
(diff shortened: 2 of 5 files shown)


fepegar:
@romainVala Can you please take a look?

fepegar:
I'm going to merge this because I accidentally started working on top of this branch and things will get messy otherwise. I think this should cover what we discussed.

torchio/data/image.py
```diff
@@ -132,3 +138,23 @@ def load(self, check_nans: bool = True) -> Tuple[torch.Tensor, np.ndarray]:
         if check_nans and torch.isnan(tensor).any():
             warnings.warn(f'NaNs found in file "{self.path}"')
         return tensor, affine
+
+    def is_2d(self) -> bool:
+        return self.shape[-3] == 1
+
+    def numpy(self) -> np.ndarray:
+        return self[DATA].numpy()
```
romainVala, line 146: may be as_numpy to be more explicit ...

torchio/data/image.py
```diff
@@ -132,3 +138,23 @@ def load(self, check_nans: bool = True) -> Tuple[torch.Tensor, np.ndarray]:
...
+    def get_center(self, lps: bool = False) -> TypeTripletFloat:
+        """Get image center in RAS (default) or LPS coordinates."""
+        image = self.as_sitk()
+        size = np.array(image.GetSize())
+        center_index = (size - 1) / 2
+        l, p, s = image.TransformContinuousIndexToPhysicalPoint(center_index)
+        if lps:
+            return (l, p, s)
+        else:
+            return (-l, -p, s)
```
romainVala, line 160: I may be a little bit confused, here: I do not understand why you should take care of RAS or LPS ?
My understanding is that one needs to define a point in world coordinate (RAS or LPS are for the anatomical coordinate system as explain here https://www.slicer.org/wiki/Coordinate_systems
But I think your right because sitk uses a LPS coordinate system ... ?

torchio/data/image.py
```diff
@@ -132,3 +138,23 @@ def load(self, check_nans: bool = True) -> Tuple[torch.Tensor, np.ndarray]:
...
+    def is_2d(self) -> bool:
+        return self.shape[-3] == 1
+
+    def numpy(self) -> np.ndarray:
+        return self[DATA].numpy()
+
+    def as_sitk(self) -> sitk.Image:
+        return nib_to_sitk(self.data, self.affine)
+
+    def get_center(self, lps: bool = False) -> TypeTripletFloat:
```
romainVala, line 151: may be a even more explicit name : 
get_matrix_center  or get_matrix_center_in_mm

torchio/transforms/augmentation/spatial/random_affine.py
```diff
@@ -125,39 +115,86 @@ def get_params(
...
+                interpolation = self.interpolation
+
+            if image.is_2d():
+                scaling_params[0] = 1
+                rotation_params[-2:] = 0
+
+            if self.use_image_center:
+                center = image.get_center(lps=True)
+            else:
+                center = None
```
romainVala, line 158: if I correctly understand 
center = [0, 0, 0] would led to the same result right ?

torchio/transforms/augmentation/spatial/random_affine.py
```diff
@@ -125,39 +115,86 @@ def get_params(
...
+                image[AFFINE],
+                scaling_params.tolist(),
+                rotation_params.tolist(),
+                interpolation,
+                center_lps=center,
+            )
+        random_parameters_dict = {
+            'scaling': scaling_params,
+            'rotation': rotation_params,
+        }
```
romainVala, line 171: here you should add the center information, so that one can reproduce the transform with the saved parameters

torchio/data/image.py
fepegar: This follows PyTorch syntax.

torchio/data/image.py
fepegar: ITK (and DICOM) assumes the first two dimensions to go left and posterior, whereas NIfTI likes right and anterior.

torchio/data/image.py
fepegar: Maybe I can write a longer docstring?

torchio/transforms/augmentation/spatial/random_affine.py
fepegar: Yes. Do you think I should make it more explicit?

torchio/transforms/augmentation/spatial/random_affine.py
fepegar: I didn't add it because it's not random, so it would be redundant. In theory, the user knows which arguments they passed to the transform.

fepegar:
About adding support to specify the center, do you think that's necessary? At the beginning I thought of that, but I thought it wasn't very useful. Is that feature something you'd be interested in?

fepegar:
(sorry I merged this before letting you finish reviewing)

romainVala:
no idea if it makes a difference at the end, but from a theoretic point of view, I do not see why one should be restricted to only zero center rotation
changing the rotation center randomly, would led to a bigger space of variation, if I had the choice I will choose the wider range,
from the torchio perspectiv, It make sense to give the user the choice ...
but I agree it is not a big deal, so not a priority ...

fepegar:
Now that there is also a `translation` parameter, I think that adds the needed variation. But point taken. If someone asks, it'll be a very easy feature to add.

Thanks @romainVala!

romainVala:
I do not know why I missed, the translation in my first review (...)
but then, yes indeed, now we have translation, I cant make an example, to apply rotation around any given point.

PR #196, opened 2020-06-20
Toch 1.1 is sufficient

### LINKED ISSUE #195: Is torch 1.2 still mandatory?
IterableDataset seems to have been removed as of #175, therefore torch>=1.1 could be sufficient. 
This is in interest of many people still relying on CUDA 9.

https://github.com/fepegar/torchio/blob/0e8172c3253050fcbf309989b57cdf9e81a3cc6c/setup.py#L20

MonsieurWave:
IterableDataset seems to have been removed as of #175, therefore torch>=1.1 could be sufficient.
This is in interest of many people still relying on CUDA 9.

resolves #195

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,7 @@ requirements = [
     'Python-Deprecated',
     'scipy',
     'SimpleITK',
-    'torch>=1.1',  # for IterableDataset
+    'torch>=1.1',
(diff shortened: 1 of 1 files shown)


fepegar:
Thanks, @MonsieurWave!

PR #197, opened 2020-06-20
Swap tensor.ndim for tensor.dim()

MonsieurWave:
Tensor.ndim is an alias for tensor.dim() but is only present in newer torch versions. 
This change would ensure backward compatibility with no impact on current behaviour.


fepegar:
Thanks, @MonsieurWave.

PR #199, opened 2020-06-20
Not applied random transform should return data untouched

### LINKED ISSUE #198: Bug: Random Transform on tensor returns Subject when not applied
**🐛Bug**
When a transform is called on a **tensor** but not applied because of its probability p, a **Subject** is returned.

**To reproduce**
```python
from torchio.transforms import RandomNoise                                                                                                                                                          
import torch                                                                                                                                                                                        
a = torch.rand(1,2,3,4)                                                                                                                                                                             
transform = RandomNoise(p=1)                                                                                                                                                                        
at1 = transform(a)                                                                                                                                                                                  
transform_p0 = RandomNoise(p=0)                                                                                                                                                                     
at2 = transform_p0(a)                                                                                                                                                                                  
type(a)                                                                                                                                                                                             
Out[9]: torch.Tensor
type(at1)                                                                                                                                                                                          
Out[10]: torch.Tensor
type(at2)                                                                                                                                                                                                                                                                                                                                                                           
Out[11]: torchio.data.subject.Subject
```
**Expected behavior**
Transform should always return same type as input

**TorchIO version** 0.16.25

**Proposed Change**
https://github.com/fepegar/torchio/blob/89b76415c2fef5b9533e68c2e7ab20fff8b8c3f7/torchio/transforms/transform.py#L54
Transform call should return data and not sample when probability is not met.
I will open a PR for this.

MonsieurWave:
fixes #198 
- When random probability is not met and transform will not be applied, return data instead of parsed sample as this transforms the input into a Subject
- small syntax fix


torchio/transforms/transform.py
```diff
@@ -51,7 +51,7 @@ def __call__(self, data: Union[Subject, torch.Tensor]):
             sample = data
         self.parse_sample(sample)
         if torch.rand(1).item() > self.probability:
-            return sample
+            return data
```
fepegar, line 54: What about just moving these two lines to the beginning of `__call__`? Then there would be no need to parse the input.

MonsieurWave: I had initially done it like that, however in that case input type errors would appear only with a certain probability making debugging harder. On the other side the parsing needlessly slows the computation...

fepegar: Yeah, that's a good point. But for debugging it wouldn't make sense to use probability less than 1 anyway, and I'd like to start optimizing things, the library can be a bit slow sometimes.

MonsieurWave: Alright, I will make the change asap.

fepegar:
Fixed in `v0.16.28`. Thanks again, Julian!

MonsieurWave:
My pleasure!

PR #207, opened 2020-06-24
Refactor k-space transforms

### LINKED ISSUE #206: RandomGhosting algorithm removes zero-frequency component
**🐛Bug**

When the ghosting algorithm removes the center of k-space, the image is too distorted.

**To reproduce**

I'm using [this image](https://www.dropbox.com/s/3342ndv600tcmjp/t1.nii.gz?dl=1) and the [TorchIO Slicer module](https://torchio.readthedocs.io/slicer.html).

8 ghosts along axis 2:

Screen Shot 2020-06-24 at 13 45 15

7 ghosts:
Screen Shot 2020-06-24 at 13 45 41

The shape of the image along axis 2 is 176, which is a multiple of 8.

**Expected behavior**

An image that looks more like the second.

**TorchIO version**

0.17.0.

fepegar:
Resolves #206.

### DIFF
--- a/tests/transforms/augmentation/test_random_motion.py
+++ b/tests/transforms/augmentation/test_random_motion.py
@@ -5,10 +5,5 @@ from ...utils import TorchioTestCase
 class TestRandomMotion(TorchioTestCase):
     """Tests for `RandomMotion`."""
     def test_random_motion(self):
-        transform = torchio.transforms.RandomMotion(
-            seed=42,
-        )
-        transformed = transform(self.sample)
-        self.sample['t2'][torchio.DATA] = self.sample['t2'][torchio.DATA] - 0.5
-        with self.assertWarns(UserWarning):
-            transformed = transform(self.sample)
+        with self.assertRaises(ValueError):
+            transform = torchio.transforms.RandomMotion(num_transforms=0)
--- a/torchio/transforms/augmentation/intensity/random_ghosting.py
+++ b/torchio/transforms/augmentation/intensity/random_ghosting.py
@@ -1,4 +1,3 @@
-import warnings
 from typing import Tuple, Optional, Union
 import torch
 import numpy as np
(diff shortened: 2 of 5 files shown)


fepegar:
@romainVala Let me know if you like the results using this implementation and I'll merge.

romainVala:
yes it looks fine to be.
but to be honest I did not use RandomGhosting so far, because this is not a common artifact for T1 image (ghosting is more present in EPI acquisitions)

fepegar:
Thanks, I didn't know that.

I meant you wanted to try the new implementation for `RandomMotion` (using the real component instead of the magnitude).

romainVala:
for RandomMotion I would not go with the real part, (although one could try to understand the difference) but the physical acquisition, is really motion in kspace and then the abs value is taken to reconstruct the images (I do not know why but all sequence always perform abs after the fft (never the real part) ...
I know from work in spectroscopy (where we do 1D fft) that both can be used, but then with difference (the peak width is shorter with the real part compare to the abs part), but this is an other story ...

may be it does not make a big difference when we go for 3D volumes, ... I do not know

PS I do not use RandomMotion either, we have an other implementation, which will be merged in torchio on day I hope (you 'll be convince if we succeed to publish it ...)

fepegar:
Well, I think only the magnitude is used because the coils are in quadrature. So if the signal for a voxel is , or , then the phase component is discarded and the result is just .

[Here's some info about this](http://mriquestions.com/real-v-imaginary.html), where this image can be found:

Screen Shot 2020-06-26 at 10 17 17

> may be it does not make a big difference when we go for 3D volumes, ... I do not know

From my experiments it seems to make [no](https://github.com/fepegar/torchio/issues/206#issuecomment-648852293) or [little](https://github.com/fepegar/torchio/issues/206#issuecomment-648931984) difference, as the imaginary component is very small.

fepegar:
Anyway, I think it's time to merge. Thanks again, @romainVala!

Issue #183, opened 2020-06-05
Queue does not take full advantage of multiprocessing


GFabien:
**Enhancement**

`Queue` has a `num_workers` attributes to take advantage of multiprocessing to load faster the samples. However, if samples are loaded in parallel, patches are extracted in the main process whatever the `num_workers` value is which can greatly augment loading time.

**To reproduce**
```python
import torch
import torchio
import timeit

image_size = 120, 120, 120
patch_size = 64, 64, 64
dataset = torchio.ImagesDataset([torchio.Subject(
    img=torchio.Image(tensor=torch.rand(image_size))
) for i in range(10)])
sampler = torchio.data.sampler.UniformSampler(patch_size)
queue = torchio.data.Queue(dataset, max_length=100, samples_per_volume=10, sampler=sampler, num_workers=10)

print(timeit.timeit(queue.fill, number=10))
```
This code yields the result `76.48914761...`

**Generating patches in workers**
The code now yields the result `10.45647265...`

My change was only to generate the patches inside the `collate_fn` of the `DataLoader`. It seems to make loading really faster but to consume more RAM (I guess because now workers that are waiting are not only waiting with a sample but also with patches). Because of this compromise I think it would be a good idea to give the user the option to get both samples and patches using multiprocess or only samples. What do you think?

Here is the change I made to the `Queue` class:
```python

    def fill(self) -> None:
        ...
        for _ in iterable:
            patches = self.get_next_subject_sample()       # CHANGE HERE
            self.patches_list.extend(patches)
        if self.shuffle_patches:
            random.shuffle(self.patches_list)

    def get_next_subject_sample(self) -> dict:
        # A StopIteration exception is expected when the queue is empty
        try:
            subject_sample = next(self.subjects_iterable)
        except StopIteration as exception:
            self.print('Queue is empty:', exception)
            self.subjects_iterable = self.get_subjects_iterable()
            subject_sample = next(self.subjects_iterable)
        return subject_sample

    def get_subjects_iterable(self) -> Iterator:
        def collate_fn(x):                      # CHANGE HERE
            iterable = self.sampler(x[0])
            return list(islice(iterable, self.samples_per_volume))
        # I need a DataLoader to handle parallelism
        # But this loader is always expected to yield single subject samples
        self.print(
            '\nCreating subjects loader with', self.num_workers, 'workers')
        subjects_loader = DataLoader(
            self.subjects_dataset,
            num_workers=self.num_workers,
[...]

PR #214, opened 2020-07-01
Optimize cropping in samplers

### LINKED ISSUE #212: Speedup crops
Hi! Thanks for your amazing library, I'll definitely cite it in my work!

**🚀 Feature**
Faster crops in GridSampler and other samplers

**Motivation**
I've tried to use GridSampler with 512x512x512 array. On my Ryzen 2700 cpu one call of __getitem__ takes ~5 seconds. I've figured out the reason: it's sitk.GetImageFromArray(array.transpose()) in nib_to_sitk function, which takes 2.5 seconds for both intensity and mask arrays. You use it on the whole source array on every apply_transform call, which is fine for smaller arrays, but becomes a huge bottleneck for big arrays.

**Pitch**
You could probably somehow refactor samplers to call nib_to_sitk once at the initialization, and then make crops from sitk array without conversion of the whole big array for every single crop

**Alternatives**
The other possible way is to get rid of sitk crop completely and use torch/numpy slicing directly

fepegar:
Related to #175, #183, 89acf63.

Fixes #212.

### DIFF
--- a/torchio/data/image.py
+++ b/torchio/data/image.py
@@ -296,3 +296,10 @@ class Image(dict):
 
     def set_check_nans(self, check_nans):
         self.check_nans = check_nans
+
+    def crop(self, index_ini, index_fin):
+        # TODO: handle affine
+        i0, j0, k0 = index_ini
+        i1, j1, k1 = index_fin
+        patch = self.data[0, i0:i1, j0:j1, k0:k1].clone()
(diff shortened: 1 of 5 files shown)


fepegar:
This cropping is now affine-aware.

Before:
Screenshot_1


After:
Screenshot_2

Issue #208, opened 2020-06-26
Reproduce a given transform


romainVala:
**🚀 Feature**
Given the history information saved in the sample for a given transform, how can I apply the exact same transform on a new volume

**Motivation**
After doing a lot of data augmentation for training and validation, I would like to visually check some specific volume (the one having the worst loss for instance).
Since saving all the transformed volume can become a very high cost (in terme of disk space) one prefer to save the transform's parameter, and have the possibility to re-create the exact same transformed volume

**Pitch**
I implemented once for the RandomElasticDeformation
```
    def apply_given_transform(self, sample: Subject, bspline_params) -> dict:
        for image_dict in sample.get_images(intensity_only=False):
            if image_dict[TYPE] == LABEL:
                interpolation = Interpolation.NEAREST
            else:
                interpolation = self.interpolation
            image_dict[DATA] = self.apply_bspline_transform(
                image_dict[DATA],
                image_dict[AFFINE],
                bspline_params,
                interpolation,
            )
        return sample

```

**Alternatives**
may be a more generic way would be to separate the apply_transform code in 2 disctinct par
one to get the random param
one to apply them
so that the apply code could be reuse in both case 
....

**Additional context**

fepegar:
I agree this would be useful, and that what you suggest makes sense. But maybe a way to do this without changing so much code everywhere is to get the current seed of the PyTorch random number generator and add it to the parameters dictionary, or rather to use a `torch.Generator` and pass put the `torch.Generator.initial_seed()` in the dictionary. I'll need to investigate this.

Useful links:
- https://pytorch.org/docs/stable/notes/randomness.html
- https://discuss.pytorch.org/t/is-there-a-randomstate-equivalent-in-pytorch-for-local-random-generator-seeding/37131
- https://pytorch.org/docs/stable/search.html?q=get_rng_state&check_keywords=yes&area=default#
- https://stackoverflow.com/a/55097968/3956024

fepegar:
I'll keep adding some findings here, for reference.

```python
 Py 3.8.2 (torchio)  ~  ipython                                                                                                                      ✔  100%  01:23  13:48:29
Python 3.8.2 (default, Mar 26 2020, 10:43:30)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.13.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import torch

In [2]: torch.save(torch.get_rng_state(), '/tmp/rng.pth')

[...]

PR #226, opened 2020-07-11
Improve random transforms reproducibility

### LINKED ISSUE #191: Seeding in Transform, make the same transform all the time
**🐛Bug**
I notice it at least for RandomAffine RandomElastic and RandomNoise

**To reproduce**
```
t = Compose([RandomNoise(seed=10)])
dataset = ImagesDataset(suj, transform=t); 

sample=dataset[0]
for i in range(0, 10):
    sample = dataset[0]
    h=sample.history
    print(h[0][1]['image']['std'])

```

Now change seed=10 to seed=None and you get a diferent number each time

I thing it is due to this line
https://github.com/fepegar/torchio/blob/0e8172c3253050fcbf309989b57cdf9e81a3cc6c/torchio/transforms/augmentation/random_transform.py#L32

it should not be called here, juste once in the init()

fepegar:
Resolves #191, #208.

I will also add docs on reproducibility, and maybe a notebook.


romainVala:
Hi
Nice PR, as most of the change is to remove code from the transform, since now save and get history is done in the subject class

I am still not sure, that to do not have the possibility to add 'normal parameter" in the history, (as it was before) is a good idea, 

Let consider random Elastic where you have to  specify the num_control_point, having only the seed saved in the history, will allow to reproduce the same transform, if and only if you put the name num_control_point
Why not saving it to the history

Other drawback of this changes, is for simple transform without random choise (like rescale or normalization), I would like to add the chosen parameter in the history too, 

but with those changes, you can now only log seed ... non ?

fepegar:
The problem with this implementation is that intensity transforms use different parameters for the different images in the `Subject`. Either we use the same for all, or we use a more complex transforms history...

> Let consider random Elastic where you have to specify the num_control_point, having only the seed saved in the history, will allow to reproduce the same transform, if and only if you put the same num_control_point
Why not saving it to the history ?

Because the user has control over what parameters they passed to the transform.

romainVala:
> Because the user has control over what parameters they passed to the transform.
yes but it makes the code more difficult (with hard coded parameter valuer), I really thing it will be more efficient to read it directly from the history (given the fact it is so simple to add !)

applying transform and reading parameters, should be dissociate

I can think at an example where you do not know : let say you want to compare the influence of num_control_point, and you define OneOf transform with 2 randomElastic call each of them with different num_control_point

romainVala:
sorry to insist, but I really like to convince you on this point 

Once you performed multiple experiment, with different transform, it is convenient to just have a look to save history in order to know what has been applied
I loose this explicit reading if we now store only the seed, but at least I can make a function that will convert the stored seed in real parameters, but for the completeness I also need the "fixed parameters"

It is also a way to check that every thing was done as planed (which is important too)

The objective is to be able to recreate a given transformed volume (because the model is very bad for this volume) why should I need to remember which num_control_point I set (or which percentiles I set for  RescaleIntensity) if those parameters can be stored in the history I have an autonomous mechanism to reproduce the given transform without prior knowledge
so I can write a generic function to reproduce any given transform (from its saved history)

don't you think it is much better so ?

If there are some drawback to store all parameters, let's make it optional, but I think it is a great feature, to make this full parameter recording possible

romainVala:
> The problem with this implementation is that intensity transforms use different parameters for the different images in the `Subject`. Either we use the same for all, or we use a more complex transforms history...

given what has been discuss here #238 we will need a more complex transforms history ... 
may be instead of having a list containing 
[ <transform_name> , transform_dict_param] 
we will have
[ <transform_name> , [<image_key>, transform_dict_param] ]
(and if the same transform is apply to all we will take the previous one)

PR #244, opened 2020-07-28
Add more tests

GFabien:
- Add tests (mainly about transforms)
- Refactor `RandomTransform.parse_range` to `Transform` and expand it to deal with ranges with specific constraints.
- Reinforce transform arguments' parsing
- Correct a problem with `Image` path handling (cropping an image without path was copying path `''` instead of `None`)
- Correct a problem with `RandomNoise` that made some tests failed in `test_transforms.py`
- Propose a minor change to `RandomSpike` (aims to be more intuitive: an `intensity` of `0` used to introduce an artifact while now an `intensity` of `0` does not produce any artifact)

### DIFF
--- a/tests/data/inference/test_grid_sampler.py
+++ b/tests/data/inference/test_grid_sampler.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 
+from copy import copy
 from torchio.data import GridSampler
 from ...utils import TorchioTestCase
 
@@ -38,3 +39,12 @@ class TestGridSampler(TorchioTestCase):
         sampler = GridSampler(self.sample, (10, 20, 30), 0)
         fixture = [[0, 0, 0, 10, 20, 30]]
         self.assertEqual(sampler.locations.tolist(), fixture)
+
+    def test_sample_shape(self):
+        patch_size = 5, 20, 20
+        patch_overlap = 2, 4, 6
+        initial_shape = copy(self.sample.shape)
+        sampler = GridSampler(
+            self.sample, patch_size, patch_overlap, padding_mode='reflect')
+        final_shape = self.sample.shape
+        self.assertEqual(initial_shape, final_shape)
--- a/tests/data/sampler/test_uniform_sampler.py
+++ b/tests/data/sampler/test_uniform_sampler.py
@@ -1,0 +1,25 @@
+import torch
+import torchio
+import numpy as np
+from torchio.data import UniformSampler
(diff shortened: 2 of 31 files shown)


GFabien:
Ok, my local torch version was accepting to have `0.0` standard deviation value in `normal_` but it is not the case for the Travis build. Any reason to choose `normal_` rather than `randn(...) * std + mean`?

fepegar:
Wow, this is impressive! You probably know the library better than me by now :D These are very helpful, and much needed changes.

One little thing. In `RandomDownsample`, the correct syntax for the Python example in the docstring for multiline statements is `...` instead of `>>>`.

> Ok, my local torch version was accepting to have 0.0 standard deviation value in normal_ but it is not the case for the Travis build. Any reason to choose normal_ rather than randn(...) * std + mean?

No reason. As long as the random parameters are computed by PyTorch (and not NumPy or Python) and results are the same, I guess it's fine for most cases.

GFabien:
> One little thing. In RandomDownsample, the correct syntax for the Python example in the docstring for multiline statements is `...` instead of `>>>`.

Ok thanks!

> No reason. As long as the random parameters are computed by PyTorch (and not NumPy or Python) and results are the same, I guess it's fine for most cases.

I haven't found much information comparing both. I know that `randn(...) * std + mean` allows the variables to be differentiable w.r.t. the mean and variance but we don't care about that here. It seems that `normal_` does not allow to take a zero standard deviation so I would rather choose `randn(...) * std + mean` but it does not change much.

fepegar:
Can you please switch to `randn` and fix the docstring? We can then merge, I think.

GFabien:
Sure, I will also add some tests about cropping images because I found a bug when cropping `ScalarImage` and `LabelMap`: type is copied to the new instance but when a type is given to one of the child classes, an error is raised.

```python
import torch
from torchio import ScalarImage

data = torch.ones((10, 10, 10))
image = ScalarImage(tensor=data)
image.crop((1, 1, 1), (5, 5, 5))
```

fepegar:
Damn! Good catch!

fepegar:
Maybe we should just drop the `type` kwarg in `v0.18`. It was useful when we used to use `dict`s, but with proper classes and inheritance, there's no point, I think.

GFabien:
I made a little change to fix it without removing `kwargs` but feel free to change it if you want!

fepegar:
Added in `v.0.17.14`. Thanks again!

PR #246, opened 2020-07-31
Add support for 4D images

### LINKED ISSUE #238: Add support for 4D images
# 🚀 Feature

Add support to read, write, sample and transform 4D images.

# Motivation

It would be convenient to support 4D images out of the box, as tensors of shape (C, S1, S2, S3) where C is the number of channels and SX are spatial dimensions (S1 = 1 for 2D images).

### Gradient directions
- Diffusion MRI

### Time

- Functional MRI
- Ultrasound sequence

### Modalities or MRI sequences

- T1, CT, US
- T1, T2, PD

### EM frequency bands

- RGB
- [Multispectral](https://www.sciencedirect.com/science/article/pii/B9780444639776000249)

### Labels

- Discrete: one-hot encoded labels
- Continuous: fractional labelmap aka fuzzy volumes aka tissue probability maps

# Pitch

Support of 4D images in:
- [x] `torchio.Image`
- [x] `torchio.io`
- [x] Samplers
- [x] Intensity transforms
- [x] Spatial transforms
- [x] Docs

# Alternatives

Pass the channels independently in different instances of `Image`.

# Additional context

#### Possible image shapes when reading

How do we know if 2D or 3D when there are 3 dimensions in the read data? Lookup table of file extensions?

##### 2D

- (height, width)
- (height, width, channels): how do we know whether the 3rd dimension is depth or channels? Easy to infer if it's 3 as probably RGB, harder otherwise (e.g. multispectral images).

##### 3D

- (height, width, depth): how do we know whether the 3rd dimension is depth or channels?
- (height, width, depth, channels): is this typical for dMRI and fMRI? What about 4D ultrasound?

### Transforms

Consider whether to apply the same or different transformations to each channel of each image, as pointed out by @romainVala in https://github.com/fepegar/torchio/issues/238#issuecomment-661708748.

## Related issues
- #100: Consider multi-channel output in the `GridAggregator`
- #213: Support 4D tensors as `torchio.Image`
- #215: Transform `torchio.Subject` - mask with multiple masks and different shapes than image
- #228: Passing mean and std to normalization transform
- #234: Inconsistent shapes for diffusion MRI
- #236: How to handle multi-modal images?
- #237: How to handle multi-class segmentation?

fepegar:
Resolves #238.


fepegar:
@romainVala the fun starts here: all transforms must work with 4D inputs so that the tests pass.

romainVala:
well done !

Issue #253, opened 2020-08-05
Update RandomLabelsToImage


fepegar:
@GFabien @romainVala 

Now that I've included the ICBM template, do you think it's a good idea to simplify a bit the example in `RandomLabelsToImage`? Also, Colin27 2008 is very large.

For example:

```python
import torchio
from torchio import RandomLabelsToImage
from torchio.datasets import ICBM2009CNonlinearSymmetryc

icbm = ICBM2009CNonlinearSymmetryc(load_4d_tissues=False)

# Default Gaussian parameters
transform = RandomLabelsToImage(
    pv_label_keys=['gm', 'wm', 'csf'],
)
transformed = transform(icbm)
transformed['image'].save('/tmp/lab2im.nii')

# Specify Gaussian parameters
gaussian_parameters = {
    'gm': dict(mean=2, std=0.1),
    'wm': dict(mean=4, std=0.1),
    'csf': dict(mean=6, std=0.1),
}
transform = RandomLabelsToImage(
    pv_label_keys=['gm', 'wm', 'csf'],
    gaussian_parameters=gaussian_parameters,
)
transformed = transform(icbm)
transformed['image'].save('/tmp/lab2im.nii')
```

![lab2im](https://user-images.githubusercontent.com/12688084/89413660-1ab23280-d721-11ea-828c-19fed918a108.png)

I can also add an option in the ICBM class to create a label map from the PV images.

GFabien:
Yes I think it would be more illustrative, there are a lot of classes in Colin27 2008 and here you can play both with PV and discretized images.

Talking about the refactoring of `RandomLabelsToImage`, what would you like to be changed? I thought you wanted to remove the `pv_label_keys` attribute to keep only the `label_key` attribute but I'm not sure anymore...

fepegar:
I haven't thought of what would need to be changed, but what it should support. How hard would it be to support passing the PV maps as a 4D image (all other transforms support 4D now). I thought this was your initial plan.

GFabien:
Yes it was indeed the initial plan but now I like having all my labels as individual attributes (it makes it very modular).

[...]

PR #261, opened 2020-08-11
Enable to give Image channels as different files

GFabien:
Make `Image` accept a sequence of paths and concatenate loaded tensors on the channel dimension. See https://github.com/fepegar/torchio/issues/253#issuecomment-669793181.

### DIFF
--- a/tests/data/test_image.py
+++ b/tests/data/test_image.py
@@ -18,7 +18,7 @@ class TestImage(TorchioTestCase):
             ScalarImage('nopath')
 
     def test_wrong_path_value(self):
-        with self.assertRaises(TypeError):
+        with self.assertRaises(RuntimeError):
             ScalarImage('~&./@#"!?X7=+')
 
     def test_wrong_path_type(self):
--- a/tests/data/test_image.py
+++ b/tests/data/test_image.py
@@ -5,7 +5,7 @@
 import copy
 import torch
 import numpy as np
-from torchio import Image, ScalarImage, LabelMap, Subject, INTENSITY, LABEL
+from torchio import ScalarImage, LabelMap, Subject, INTENSITY, LABEL, STEM
 from ..utils import TorchioTestCase
 from torchio import RandomFlip, RandomAffine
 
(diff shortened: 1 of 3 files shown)


torchio/data/image.py
```diff
@@ -124,7 +126,7 @@ class Image(dict):
     """
     def __init__(
             self,
-            path: Optional[TypePath] = None,
+            path: Union[TypePath, Sequence[TypePath]] = None,
```
fepegar, line 129: If `path` can be `None`, this should be `Optional[TypePath, Union[TypePath, Sequence[TypePath]]]` or `Union[TypePath, Sequence[TypePath], None]`. I think the latter is more readable.

torchio/data/image.py
```diff
@@ -262,16 +264,35 @@ def get_bounds(self):
...
+                path = Path(path).expanduser()
+            except TypeError:
+                message = f'Conversion to path not possible for variable: ' \
+                          f'{path}'
+                raise TypeError(message)
+        else:
+            try:
+                path = [Path(p).expanduser() for p in path]
+            except TypeError:
+                message = f'Conversion to list of paths not possible for ' \
```
fepegar, line 284: Can you use brackets for consistency?

```python
message = (
    'sdfs'
    f'sdfl{sd}'
)
```

torchio/data/image.py
```diff
@@ -262,16 +264,35 @@ def get_bounds(self):
...
+                          f'variable: {path}'
+                raise TypeError(message)
+
+        if isinstance(path, Path):
```
fepegar, line 288: Can the input be a list of dirs?

torchio/data/image.py
```diff
@@ -309,11 +330,35 @@ def _load(self) -> Tuple[torch.Tensor, np.ndarray]:
...
+                warnings.warn(message, RuntimeWarning)
+
+            if not tensor.shape[1:] == new_tensor.shape[1:]:
+                message = f'Files shape do not match, found {tensor.shape}' \
+                          f'and {new_tensor.shape}'
+                RuntimeError(message)
+
+            tensors.append(new_tensor)
+
+        tensor = torch.cat(tensors, dim=0)
```
fepegar, line 360: I think you can omit `dim=0`.

torchio/utils.py
```diff
@@ -43,12 +43,16 @@ def to_tuple(
...
+        ) -> Union[str, List[str]]:
     """
     '/home/user/image.nii.gz' -> 'image'
     """
-    path = Path(path)
-    return path.name.split('.')[0]
+    if isinstance(path, (str, Path)):
+        path = Path(path)
+        return path.name.split('.')[0]
+    return [Path(p).name.split('.')[0] for p in path]
```
fepegar, line 55: Is there a way to not repeat this code? Maybe leave this as it was and write a new `get_stems`? Or an inner function `_get_stem` that returns `path.name.split('.')[0]`.

torchio/data/image.py
GFabien: I don't know, it seems that for a single path a dir was acceptable, why not for a list of paths? Even if it corresponds to images with several channels it will work as long as they have the same spatial dimensions.

torchio/data/image.py
GFabien: Maybe, but isn't it better if it's explicit?

torchio/data/image.py
fepegar: It's just the impression I got by skimming through the code here. I'll get a better sense of it once I've properly gone through it.

torchio/data/image.py
fepegar: I don't think it helps in this case, as it is the default. It's like using `print(s, sep='\n')` instead of `print(s)`.

tests/data/test_image.py
```diff
@@ -17,6 +17,10 @@ def test_image_not_found(self):
         with self.assertRaises(FileNotFoundError):
             ScalarImage('nopath')
 
+    def test_wrong_path_value(self):
+        with self.assertRaises(TypeError):
+            ScalarImage('~&./@#"!?X7=+')
```
fepegar, line 22: Looks like Obélix swearing haha.

However I'm not sure this should raise a `TypeError`. The `type` is `str`, which is fine. Why not `ValueError`?

GFabien: Yes I got inspired for this one haha.

The caught one is a `RuntimeError`, do you prefer to have a `ValueError` or a `RuntimeError` then?

fepegar: I think propagating that `RuntimeError` is the wise choice.

fepegar:
Thanks, @GFabien! Added in `v0.17.26`.

PR #264, opened 2020-08-12
Refactor RandomLabelsToImage

GFabien:
Refactor `RandomLabelsToImage` as discussed in https://github.com/fepegar/torchio/issues/253.

Examples have been rewritten using the ICBM template. Here are some images created using the examples:
- Using default parameters:
Figure_1
- Using `mean=[0.33, 0.66, 1]` and `std=[0, 0, 0]`:
Figure_2
- Same as before but using `discretize=True`:
Figure_3
- Blurring precedent result:
Figure_4
- Filling CSF and background with T1 image:
Figure_5

### DIFF
--- a/torchio/transforms/augmentation/intensity/random_labels_to_image.py
+++ b/torchio/transforms/augmentation/intensity/random_labels_to_image.py
@@ -1,15 +1,12 @@
 
-from typing import Union, Tuple, Optional, Dict, Sequence, List
+from typing import Tuple, Optional, Sequence, List
 import torch
-import numpy as np
-from ....torchio import DATA, TypeData, TypeRangeFloat, TypeNumber, AFFINE
+from ....torchio import DATA, AFFINE, TypeData, TypeRangeFloat
+from ....utils import check_sequence
 from ....data.subject import Subject
 from ....data.image import ScalarImage
 from ... import IntensityTransform
 from .. import RandomTransform
-
-
-TypeGaussian = Optional[Dict[Union[str, TypeNumber], Dict[str, TypeRangeFloat]]]
 
 
 class RandomLabelsToImage(RandomTransform, IntensityTransform):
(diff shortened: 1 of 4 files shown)


fepegar:
Beautiful! I wonder if there's a way to improve the default parameters so that means are more varied. For example, if no defaults are given, the means could be uniformly distributed between e.g. 0 and 1 and randomly assigned to the labels, and sth similar with the stds. What do you think?

fepegar:
Has the 4D support made the usage simpler?

torchio/transforms/augmentation/intensity/random_labels_to_image.py
```diff
@@ -51,38 +62,39 @@ class RandomLabelsToImage(RandomTransform, IntensityTransform):
...
-        >>> transformed = transform(colin)  # colin's key 't1' has been replaced with the simulated image
+        >>> transformed = transform(sample)  # sample's key 't1' has been replaced with the simulated image
     """
     def __init__(
             self,
-            label_key: Optional[str] = None,
-            pv_label_keys: Optional[Sequence[str]] = None,
+            label_key: str,
+            used_labels: Optional[Sequence[int]] = None,
             image_key: str = 'image',
```
fepegar, line 95: Can we use something less likely to have already been used by the user? For example, `image_from_labels`.

GFabien:
I think it's almost already the case. For example, the default range for the means is `[0.1, 0.9]` so if you don't precise anything, for each label, the associated mean will be sampled uniformly between 0.1 and 0.9. We could still change the default range to be `[0, 1]` but I'm not sure it's useful.

GFabien:
> Has the 4D support made the usage simpler?

Yes definitely, but what made it really simpler was to say that we could not have several 3D labels anymore! The only "tricky" thing is that you have to define the right number of means and stds and if you want to set one you have to set them all but I think that's ok.

fepegar:
> I think it's almost already the case. For example, the default range for the means is [0.1, 0.9] so if you don't precise anything, for each label, the associated mean will be sampled uniformly between 0.1 and 0.9. We could still change the default range to be [0, 1] but I'm not sure it's useful.

Strictly speaking, I meant that the means could be uniformly distributed between 0 and 1, not sampled from a uniform distribution between 0 and 1.

So e.g. shuffle(torch.linspace(0, 1, 6)) = 0.2, 0.0, 0.8, 1 and 0.6, not torch.rand(6) = tensor([0.7368, 0.1984, 0.3228, 0.7891, 0.4516, 0.5674]). Does that make sense?

GFabien:
>Strictly speaking, I meant that the means could be uniformly distributed between 0 and 1, not sampled from a uniform distribution between 0 and 1.

Oh OK, I get it now! From a training point of view I would say that your suggestion will lead to "predefined" mean values that could be annoying for the model to learn something more general but from a visualization point of view it would definitely be better.

fepegar:
You're right it's fine as it is now.

fepegar:
Thanks again! Added in `v0.17.27`.

PR #272, opened 2020-08-17
Use SimpleITK to resample

### LINKED ISSUE #227: Use SimpleITK to resample
**🚀 Feature**

Currently, resampling in `Resample` is done using NiBabel, which performs interpolation using B-splines. Using a spline of order 0 and 1, one gets the equivalent of nearest neighbor and linear interpolation. However, both take as long. With `SimpleITK` the range of different interpolation techniques is larger and times between e.g. nearest and linear would be very different, which is desired.

Furthermore, the choices for interpolation techniques would be consistent with the rest of the transforms.

Finally, this is a step further towards having a unified transform for `RandomAffine`, `RandomDownsample` and `RandomElasticDeformation`.

### LINKED ISSUE #271: Resample transform reorients to RAS
```python
In [1]: import torchio as tio

In [2]: fpg = tio.datasets.FPG()

In [4]: fpg.t1.orientation
Out[4]: ('P', 'I', 'R')

In [5]: res = tio.Resample(4)(fpg)

In [6]: res.t1.orientation
Out[6]: ('R', 'A', 'S')
```

Related to #227.

fepegar:
Fixes #271.
Resolves #227.

# Benchmark with Colin 1998 T1 (1 mm iso)

## Upsample (0.87 mm iso)
Old nearest: **2.35 s ± 360 ms**
Old linear: 2.9 s ± 567 ms
New nearest: **326 ms ± 13.8 ms**
New linear: 561 ms ± 233 ms

## Downsample (1.87 mm iso)
Old nearest: 236 ms ± 15.3 ms
Old linear: 335 ms ± 58.9 ms
New nearest: 236 ms ± 12.7 ms
New linear: 244 ms ± 12.1 ms

### DIFF
--- a/tests/transforms/preprocessing/test_resample.py
+++ b/tests/transforms/preprocessing/test_resample.py
@@ -32,13 +32,13 @@ class TestResample(TorchioTestCase):
         affine_name = 'pre_affine'
         transform = Resample(spacing, pre_affine_name=affine_name)
         transformed = transform(self.sample)
-        for image_dict in transformed.values():
-            if affine_name in image_dict.keys():
+        for image in transformed.values():
+            if affine_name in image:
                 new_affine = np.eye(4)
                 new_affine[0, 3] = 10
-                assert_array_equal(image_dict[AFFINE], new_affine)
+                assert_array_equal(image[AFFINE], new_affine)
             else:
-                assert_array_equal(image_dict[AFFINE], np.eye(4))
+                assert_array_equal(image[AFFINE], np.eye(4))
 
     def test_missing_affine(self):
         transform = Resample(1, pre_affine_name='missing')
@@ -69,3 +69,8 @@ class TestResample(TorchioTestCase):
         transform = Resample('missing')
         with self.assertRaises(ValueError):
             transform(self.sample)
+
+    def test_2d(self):
+        sample = self.make_2d(self.sample)
+        transform = Resample(2)
+        transform(sample)
--- a/tests/transforms/preprocessing/test_resample.py
+++ b/tests/transforms/preprocessing/test_resample.py
@@ -1,5 +1,5 @@
 import numpy as np
-from numpy.testing import assert_array_equal
+from numpy.testing import assert_array_equal, assert_array_almost_equal
 from torchio import DATA, AFFINE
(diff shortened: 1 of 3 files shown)


fepegar:
@GFabien could you please take a look at `test_affine` in `test_resample.py`? I'm not sure it's correct.

GFabien:
>@GFabien could you please take a look at test_affine in test_resample.py? I'm not sure it's correct.

I think it is correct, with the pre affine matrix you update the affine and resample using the updated affine. In the test, the affine is the identity, the pre affine matrix is
```python
np.array([
    [1,   0,   0,  10],
    [0,   1,   0,   0],
    [0,   0,  1.2,  0],
    [0,   0,   0,   1]
])
```
so the updated affine is equal to the pre affine matrix. The transform is supposed to resample the image to spacing `(1, 1, 1)` so we should end up with the following affine
```python
np.array([
    [1,   0,   0,  10],
    [0,   1,   0,   0],
    [0,   0,   1,   0],
    [0,   0,   0,   1]
])
```
or am I missing something? ITK finds the following one:
```python
np.array([
    [1,   0,   0,   10],
    [0,   1,   0,    0],
    [0,   0,   1, -0.1],
    [0,   0,   0,    1]
])
```

fepegar:
Hmm I think I misinterpreted something in the test last night. Anyway, you're right.

However, a small displacement of the origin is expected if we want to keep exactly the same bounding box so I think the fixture should include that -0.1.

Think of a little square and a big square representing the first pixel (0, 0) of an image at 1 and 2 mm spacing. If both pixels are centered (image origin) at the same position, the bounds of the image will be different. To keep it constant, the small square needs to be a bit different. In this case, the difference is only in the z-axis because of the 1.2. Does any of that make sense?

GFabien:
Yes it does, I think you're right. I just tried several spacing to check it and it seems to always generate a displacement of half the difference of spacings which is consistent with your statement. So yes, the fixture should take this -0.1 into account.

fepegar:
FYI the origin displacement is computed in lines 208 and 209 of resample.py.

Thanks for checking that test! Maybe NiBabel doesn't change the origin and that's why it was passing. It's just an implementation detail, anyway.

PR #274, opened 2020-08-18
Fix check of spatial shape consistency in weighted sampler

OeslleLucena:


fepegar:
Fixed in `v0.17.29`. Thanks for you contribution, @OeslleLucena :D

fepegar:
@all-contributors please add @OeslleLucena for bug

PR #287, opened 2020-08-19
Adding transform reproducibility

GReguig:
Hello,

after the issue #208, I tried to use a rather generic way to reproduce a transform from a subject's history. The main idea is using a seed for each transform call, which was added to the base transform abstract class. 

Also, the seed (whether the transform is Random or not) is saved in the history. This allows transforms which originally do not save anything in the subject's history to at least save something in the history.

The reproducibility comes from the compose_from_history function in torchio/transforms/augmentation/composition.py which returns a Compose object and the seeds of each of its transforms to reproduce the given history. 

Also, I added a few tests which have all passed but it could certainly be more exhaustive. The testing file was inspired by @fepegar 's branch on reproducibility.

The code is not quite clean yet (for example, I commented all the add_transform calls in the transform to avoid interference in the history, but I did not clean the whole transform code). 

Completely open to criticism !

Thanks for reading !


GReguig:
Hello @fepegar,

thank you for reviewing and sorry that I took so long to push a new version. I tried fixing everything by changing as little code as possible. Most of the changes come from the Compose and Transform class where the seeds are handled differently. I had to add a (kind of) special case for the Compose, which was a little bit tricky. 

Also I added a test checking for the rng state before and after. Still, the code is not quite clean yet, I am waiting to have a fully functional reproducibility pattern to clean every transform's code. 

Again, completely open to criticism !

fepegar:
Hi, @GReguig. Sorry for the delay. This looks pretty good! But I'm struggling to review it, mostly because of the conflicts. I'll try to fix some, then add some comments in the code.

It'd be great if you could explain a bit, with plain words, how the new system works.

torchio/transforms/augmentation/composition.py
```diff
@@ -20,12 +22,28 @@ class Compose(Transform):
     .. note::
         This is a thin wrapper of :py:class:`torchvision.transforms.Compose`.
     """
-    def __init__(self, transforms: Sequence[Transform], p: float = 1):
```
fepegar, line 23: Why is `transforms` this a kwarg now?

torchio/transforms/augmentation/composition.py
```diff
@@ -20,12 +22,28 @@ class Compose(Transform):
...
+            transforms: Optional[Sequence[Transform]] = None,
+            p: float = 1,
+            ):
         super().__init__(p=p)
+        transforms = [] if transforms is None else transforms
         self.transform = PyTorchCompose(transforms)
 
-    def apply_transform(self, subject: Subject):
-        return self.transform(subject)
+    def __call__(self, data: Union[Subject, torch.Tensor, np.ndarray], seeds: List = None):
```
fepegar, line 34: Is this compatible with e.g. a Compose or OneOf within a Compose?

torchio/transforms/augmentation/composition.py
```diff
@@ -106,3 +124,21 @@ def _normalize_probabilities(transforms_dict: dict):
...
+
+def compose_from_history(history: List):
+    """
+    Builds a composition of transformations from a given subject history
+    :param history: subject history given as a list of tuples containing (transformation_name, transformation_parameters)
+    :return: Tuple (Compose of transforms, list of seeds to reproduce the transforms from the history)
+    """
+    trsfm_list = []
+    seed_list = []
+    for trsfm_history in history:
```
fepegar, line 137: Can you switch to `for trsfm_name, trsfm_params in trsfm_history`?

torchio/transforms/augmentation/composition.py
```diff
@@ -106,3 +124,21 @@ def _normalize_probabilities(transforms_dict: dict):
...
         for transform, probability in transforms_dict.items():
             transforms_dict[transform] = probability / probabilities.sum()
+
+
+def compose_from_history(history: List):
+    """
+    Builds a composition of transformations from a given subject history
+    :param history: subject history given as a list of tuples containing (transformation_name, transformation_parameters)
+    :return: Tuple (Compose of transforms, list of seeds to reproduce the transforms from the history)
+    """
```
fepegar, line 134: Can you switch to Google style instead of NumPy style?

https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html

torchio/transforms/transform.py
```diff
@@ -50,8 +50,9 @@ def __init__(
         self.copy = copy
         self.keys = keys
         self.default_image_name = 'default_image_name'
+        self.transform_params = {}
 
-    def __call__(self, data: TypeTransformInput) -> TypeTransformInput:
+    def __call__(self, data: Union[Subject, torch.Tensor, np.ndarray], seed: Union[List[int], int]=None):
```
fepegar, line 55: Can you add `seed` in the docstring?

torchio/transforms/transform.py
```diff
@@ -62,7 +63,22 @@ def __call__(self, data: TypeTransformInput) -> TypeTransformInput:
...
+            seed = gen_seed()
+
+        #Store the current rng_state to reset it after the execution
+        torch_rng_state = torch.random.get_rng_state()
+        if isinstance(seed, int):
+            torch.manual_seed(seed=seed)
+
+        self.transform_params = {}
+        self._store_params()
+        self.transform_params["seed"] = seed
```
fepegar, line 77: Can the seed stuff be moved into `RandomTransform` instead?

torchio/utils.py
```diff
@@ -316,6 +318,22 @@ def check_sequence(sequence: Sequence, name: str):
         raise TypeError(message)
 
 
+def gen_seed():
+    """
+    Random seed generator to avoid overflow
```
fepegar, line 323: As above: can you stick to Google docstrings style?

fepegar (changes requested):
I have added some comments. Can you please take a look, @GReguig?

fepegar:
Btw, can you please run `pip install -r requirements-dev.txt && pre-commit install && pre-commit run --all-files`?

fepegar:
I'm going to fix the flake8 errors shown after running `pre-commit run --all-files`.

torchio/transforms/transform.py
romainVala: not sure, there are 2 functionality here : handeling the seed, which is only related to randomTransform but also store transform param, with _store_params which then need to be in transform.py

I let ghiles explain how the seeding works, but for the parameter storing it simply add all parameters of the class. This make it convenient to add any require parameter, 
for instance in the random_affine.py I just had to add:
        self.apply_scales =  scaling_params
        self.apply_rot = rotation_params
        self.apply_trans = translation_params
(and now we know it is not enough I could add
self.afffine = ...

I know we did not agreed on previous discussion on which parameters needs to be include in the history, but at least with this mechanism, it make it easy to modify torchio transform to add any need information ... non ?

fepegar:
If you're unable to commit due to `pre-commit`, just use `git commit --no-verify`.

torchio/transforms/transform.py
```diff
@@ -62,7 +74,22 @@ def __call__(self, data: TypeTransformInput) -> TypeTransformInput:
...
+        torch_rng_state = torch.random.get_rng_state()
+        if isinstance(seed, int):
+            torch.manual_seed(seed=seed)
+
+        self.transform_params = {}
+        self._store_params()
+        self.transform_params['seed'] = seed
+
         if torch.rand(1).item() > self.probability:
+            # If not a Compose
```
fepegar, line 90: What if it is `OneOf`?

And why is the transform added to the history if it's not applied to the subject?

fepegar (changes requested):

torchio/transforms/transform.py
fepegar: Hmm ok, I think I see why it's not possible to move to `RandomTransform`. I hope we'll be able to come up with cleaner solutions for this soon.

romainVala: may be the seeding part to random transform and _store_params() to transform.py .. @GReguig ?

GReguig: It is clearly possible to do as @romainVala is suggesting, treating the seeding part in the 'RandomTransform' part and executing the '_store_params' in the 'Transform'. However, it introduces an inconsistency in the different types of transforms and will force very specific treatments when wanting to loop over a sequence of transforms. This would be problematic if we want to reproduce a whole Compose, some transforms will have a seed and others won't. 

I personally think it would be best to keep the consistency (with a seed for every transform, even if it is not a random one) rather than handling very specific treatments afterwards. What do you think ?

torchio/transforms/transform.py
GReguig: I thought it would still be useful to keep the information that a transform was not executed because of the random probability, but I cannot think of a specific case where it would be, so I will delete this one.

I am currently working on handling the `OneOf` case, I will update this PR as soon as I get it working

torchio/transforms/augmentation/composition.py
GReguig: I set a default value for `transforms` in order to keep the generic treatment of the reproducibility. In this implementation, the transform is initialized without any argument before setting the arguments manually using the `__dict__` attribute of the class. Since the `Compose` transform could not be initialized without arguments, I used this "trick" to not have a specific code for this class. I could try to find another way if you prefer

torchio/transforms/transform.py
fepegar: I just think it doesn't make a lot of sense to store the seed for a transform that is not random. I'm also not sure that the effort to reproduce a composed transform is necessary. You could just loop over the history and applied one transform after the other, right?

torchio/transforms/transform.py
fepegar: As I said above, maybe there's no need to work on reproducibility for `Compose` and `OneOf`. Iterating over the history should do.

GReguig: Clearly, yes...

romainVala: Yes, and it should not be store in the history either this will make the history cleaner than we have now

torchio/transforms/transform.py
romainVala: I agree, 
it may then worth to add the specific treatments in the code, in order to get a lighter (and then cleaner history)

GReguig:
Hello,

I think I covered all the requested changes with this last push. I moved the seeding part to the `RandomTransform` and covered some special cases for the `compose_from_history`:

- the first one is the `interpolation` argument (as used in `AffineTransform` for example) which required a specific conversion. I  also changed the name of the interpolator attribute in the `RandomElasticDeformation` from `image_interpolation` to `interpolation` to keep consistency of the interpolation object for all Transforms.

- the second one is the case of the `patch_size` argument from `RandomSwap` which was not converted to a list of int but was kept as a string, therefore I added an ugly special case using `json.loads`. I could not think of a cleaner way...

- the third one was about the transforms requiring an argument for their initialisation (`RandomLabelsToImage`,`Resample`). The arguments are extracted from the history and used to initialise the transforms.

- finally, the last special case is about the `RandomDownsample` transform which uses a `Resample` transform. As the `Resample` transform appears in the history, there is no need to use the `RandomDownsample`.

Tell me if I forgot something or if there is something else to improve/add.

romainVala:
as I said, there are 2 distinct part in this PR, the seeding mechanism, and the parameter recall
about the second one, @fepegar : do you thing we should make something more explicit, following what you propose here https://github.com/fepegar/torchio/issues/299#issuecomment-718796240

fepegar:
> @fepegar : do you think we should make something more explicit, following what you propose here

Yes, but let's merge this for now. Many thanks, @GReguig and @romainVala!

Issue #232, opened 2020-07-16
Request for new intensity augmentation options


sarthakpati:
**🚀 Feature**
It would be great if the following options could be added:
- [Color augmentations](https://github.com/MIC-DKFZ/batchgenerators#supported-augmentations)
  - brightness (additive, multiplivative)
  - contrast
  - gamma (like gamma correction in photo editing)

**Motivation**

These would take care of cases where the input data is not great.

**Pitch**

The aforementioned options would make DL models more robust.

**Alternatives**

Using other packages.

romainVala:
torchio is currently working only with gray level images.

You can change the intensity dynamic, with the RescaleIntensity transform or histogram Normalization (but they are currently more for data homogenization than for data augmentation) 

I think it is a good idea to add those transform

theabster:
@romainVala I'm not quite sure I understand. I have 3D images of proteins saved as TIFF files. The samples have all three color channels for each "slice" of the image. So each sample has dimensions of something like (40, 512, 512, 3) for a sample with 40 slices. Would torchio not work with this? Would the same transform be applied to all 3 channels implicitly, or do I need to do this myself somehow?

romainVala:
Hello
torchio was developed with MRI data in mind, multi channel is often use to concatenate several modalities, but I have no experience with it.

From a quick look at the code, I am not sure for spatial transform @fepegar how are multiple channel handle ? `assert len(tensor) == 1` is not a restriction to single channel ?

For intensity transform, (scalling or znormalization), the statistics are compute from a single array (that contain all the channel)
would be interesting to have an option to handle channels separately 
I guess for color channel, it is a special case, and there is a need for special transforms

Note that it is quite easy to build a new transform for specific use. ... if on wants to give a try and propose a PR you are wellcome ...

fepegar:
@sarthakpati 

> It would be great if the following options could be added:
Color augmentations
brightness (additive, multiplivative)
contrast
gamma (like gamma correction in photo editing)

As I said in #231, would you like to contribute?

The utility of some of these for medical image processing is not very clear. For example, pixel intensities in MRI are often quite arbitrary, unlike in computer vision (0-255), so there's not much point in adding (brightness) or multiplying (contrast) intensities. 

You can use the [`Lambda`](https://torchio.readthedocs.io/transforms/others.html#lambda) to easily implement custom transforms.

---

@theabster 

[...]

PR #290, opened 2020-08-25
Add RandomGamma transform

GFabien:
Add `RandomGamma` as discussed in https://github.com/fepegar/torchio/issues/232.

Here are 3 images using `log_gamma_range = (-0.5, -0.5)`, `log_gamma_range = (0, 0)` and `log_gamma_range = (0.5, 0.5)` on `torchio.datasets.FPG`:
low_gamma
no_gamma
high_gamma

### DIFF
--- a/tests/transforms/augmentation/test_random_gamma.py
+++ b/tests/transforms/augmentation/test_random_gamma.py
@@ -1,0 +1,31 @@
+from torchio import RandomGamma
+from ...utils import TorchioTestCase
+from numpy.testing import assert_array_equal
+
+
+class TestRandomGamma(TorchioTestCase):
+    """Tests for `RandomGamma`."""
+    def test_with_zero_gamma(self):
+        transform = RandomGamma(log_gamma_range=0)
+        transformed = transform(self.sample)
+        assert_array_equal(self.sample.t1.data, transformed.t1.data)
+
+    def test_with_non_zero_gamma(self):
+        transform = RandomGamma(log_gamma_range=(0.1, 0.3))
+        transformed = transform(self.sample)
+        with self.assertRaises(AssertionError):
+            assert_array_equal(self.sample.t1.data, transformed.t1.data)
+
+    def test_with_high_gamma(self):
+        transform = RandomGamma(log_gamma_range=(100, 100))
+        transformed = transform(self.sample)
+        assert_array_equal(self.sample.t1.data == 1, transformed.t1.data)
+
+    def test_with_low_gamma(self):
+        transform = RandomGamma(log_gamma_range=(-100, -100))
+        transformed = transform(self.sample)
+        assert_array_equal(self.sample.t1.data > 0, transformed.t1.data)
+
+    def test_wrong_gamma_type(self):
+        with self.assertRaises(ValueError):
+            RandomGamma(log_gamma_range='wrong')
--- a/tests/transforms/augmentation/test_random_gamma.py
+++ b/tests/transforms/augmentation/test_random_gamma.py
@@ -1,6 +1,5 @@
 from torchio import RandomGamma
 from ...utils import TorchioTestCase
-from numpy.testing import assert_array_almost_equal
 
 
 class TestRandomGamma(TorchioTestCase):
@@ -8,23 +7,26 @@ class TestRandomGamma(TorchioTestCase):
     def test_with_zero_gamma(self):
         transform = RandomGamma(log_gamma_range=0)
         transformed = transform(self.sample)
-        assert_array_almost_equal(self.sample.t1.data, transformed.t1.data)
+        self.assertTensorAlmostEqual(self.sample.t1.data, transformed.t1.data)
 
     def test_with_non_zero_gamma(self):
         transform = RandomGamma(log_gamma_range=(0.1, 0.3))
         transformed = transform(self.sample)
-        with self.assertRaises(AssertionError):
-            assert_array_almost_equal(self.sample.t1.data, transformed.t1.data)
+        self.assertTensorNotEqual(self.sample.t1.data, transformed.t1.data)
 
     def test_with_high_gamma(self):
         transform = RandomGamma(log_gamma_range=(100, 100))
         transformed = transform(self.sample)
-        assert_array_almost_equal(self.sample.t1.data == 1, transformed.t1.data)
+        self.assertTensorAlmostEqual(
+            self.sample.t1.data == 1, transformed.t1.data
+        )
 
     def test_with_low_gamma(self):
         transform = RandomGamma(log_gamma_range=(-100, -100))
         transformed = transform(self.sample)
-        assert_array_almost_equal(self.sample.t1.data > 0, transformed.t1.data)
+        self.assertTensorAlmostEqual(
+            self.sample.t1.data > 0, transformed.t1.data
+        )
 
     def test_wrong_gamma_type(self):
         with self.assertRaises(ValueError):
--- a/tests/transforms/augmentation/test_random_gamma.py
+++ b/tests/transforms/augmentation/test_random_gamma.py
@@ -5,24 +5,24 @@ from ...utils import TorchioTestCase
 class TestRandomGamma(TorchioTestCase):
     """Tests for `RandomGamma`."""
     def test_with_zero_gamma(self):
-        transform = RandomGamma(log_gamma_range=0)
+        transform = RandomGamma(log_gamma=0)
         transformed = transform(self.sample)
         self.assertTensorAlmostEqual(self.sample.t1.data, transformed.t1.data)
 
     def test_with_non_zero_gamma(self):
-        transform = RandomGamma(log_gamma_range=(0.1, 0.3))
+        transform = RandomGamma(log_gamma=(0.1, 0.3))
         transformed = transform(self.sample)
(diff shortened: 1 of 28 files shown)


tests/transforms/augmentation/test_random_gamma.py
```diff
@@ -0,0 +1,31 @@
...
+from ...utils import TorchioTestCase
+from numpy.testing import assert_array_equal
+
+
+class TestRandomGamma(TorchioTestCase):
+    """Tests for `RandomGamma`."""
+    def test_with_zero_gamma(self):
+        transform = RandomGamma(log_gamma_range=0)
+        transformed = transform(self.sample)
+        assert_array_equal(self.sample.t1.data, transformed.t1.data)
```
fepegar, line 11: If an operation will be performed, we should use `assert_array_almost_equal`. Otherwise, we can detect that the input is `0` and do nothing (see #292).

tests/transforms/augmentation/test_random_gamma.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def test_with_high_gamma(self):
+        transform = RandomGamma(log_gamma_range=(100, 100))
+        transformed = transform(self.sample)
+        assert_array_equal(self.sample.t1.data == 1, transformed.t1.data)
+
+    def test_with_low_gamma(self):
+        transform = RandomGamma(log_gamma_range=(-100, -100))
+        transformed = transform(self.sample)
+        assert_array_equal(self.sample.t1.data > 0, transformed.t1.data)
```
fepegar, line 27: Can you please explain these?

torchio/transforms/__init__.py
```diff
@@ -23,6 +23,7 @@
 from .augmentation.intensity import RandomGhosting
 from .augmentation.intensity import RandomBiasField
 from .augmentation.intensity import RandomLabelsToImage
+from .augmentation.intensity import RandomGamma
```
fepegar, line 26: https://twitter.com/GIStuart/status/1287713877422215169?s=20

:D

torchio/transforms/augmentation/intensity/random_gamma.py
```diff
@@ -0,0 +1,56 @@
...
+    Example:
+        >>> import torchio
+        >>> from torchio import RandomGamma, RescaleIntensity, Compose
+        >>> from torchio.datasets import FPG
+        >>> sample = FPG()
+        >>> gamma_transform = RandomGamma(log_gamma_range=(-0.3, 0.3))
+        >>> # It's better to rescale data to [0, 1] to avoid exploding values
+        >>> rescale_transform = RescaleIntensity((0, 1), (1, 99))
+        >>> transform = Compose([rescale_transform, gamma_transform])
+        >>> transformed = transform(sample)
```
fepegar, line 31: Can you add the transform to `docs/source/transforms/augmentation.rst`? It'll make it easier to check the docstring.

torchio/transforms/augmentation/intensity/random_gamma.py
```diff
@@ -0,0 +1,56 @@
+from typing import Tuple, Optional, List
+import torch
+from ....torchio import DATA, TypeRangeFloat
+from ....data.subject import Subject
+from .. import RandomTransform
+
+
+class RandomGamma(RandomTransform):
+    r"""Change contrast of an image by setting its values to the power
```
fepegar, line 9: I'm not sure "set a value to the power x" makes sense to me. What about "raising" instead of "setting"?

torchio/transforms/augmentation/intensity/random_gamma.py
```diff
@@ -0,0 +1,56 @@
...
+    :math:`\gamma`.
+
+    Args:
+        log_gamma_range: Tuple :math:`(a, b)` to compute the factor
```
fepegar, line 13: We haven't used `_range` in other transforms where a range is expected.

I feel like passing the range of log makes more sense than a standard range, but can you please explain this?

torchio/transforms/augmentation/intensity/random_gamma.py
```diff
@@ -0,0 +1,56 @@
...
+        keys: See :py:class:`~torchio.transforms.Transform`.
+
+    Example:
+        >>> import torchio
+        >>> from torchio import RandomGamma, RescaleIntensity, Compose
+        >>> from torchio.datasets import FPG
+        >>> sample = FPG()
+        >>> gamma_transform = RandomGamma(log_gamma_range=(-0.3, 0.3))
+        >>> # It's better to rescale data to [0, 1] to avoid exploding values
+        >>> rescale_transform = RescaleIntensity((0, 1), (1, 99))
```
fepegar, line 29: Can you add the missing kwarg?

torchio/transforms/augmentation/intensity/random_gamma.py
```diff
@@ -0,0 +1,56 @@
...
+        seed: See :py:class:`~torchio.transforms.augmentation.RandomTransform`.
+        keys: See :py:class:`~torchio.transforms.Transform`.
+
+    Example:
+        >>> import torchio
+        >>> from torchio import RandomGamma, RescaleIntensity, Compose
+        >>> from torchio.datasets import FPG
+        >>> sample = FPG()
+        >>> gamma_transform = RandomGamma(log_gamma_range=(-0.3, 0.3))
+        >>> # It's better to rescale data to [0, 1] to avoid exploding values
```
fepegar, line 28: Can this be elaborated in a `..note` in the docstring?

fepegar (changes requested):
Thanks! I've added some comments.

tests/transforms/augmentation/test_random_gamma.py
GFabien: It is related to the `..note` I will have to add but as we give a log range, the power at which we raise the intensity can be quickly very low or very high pushing towards extreme values. Therefore, if we set `log_gamma_range=(100, 100)` with intensity values between 0 and 1, all non zero values will be pushed to 1 when raised to the power gamma. It is just a test to make sure it behaves well regarding high or low values of gamma.

torchio/transforms/augmentation/intensity/random_gamma.py
GFabien: I think it's mainly a design choice. Again it's related to the `..note` but to avoid numerical instability, one needs that `x ** gamma` does not go towards infinity, this is ensured only if x is in [0, 1] and `gamma >= 0`. Therefore passing a range of log allows to always have `gamma > 0` (`gamma = 0` is not interesting).

torchio/transforms/augmentation/intensity/random_gamma.py
GFabien: Do you mean writing `rescale_transform = RescaleIntensity(out_min_max=(0, 1), percentiles=(1, 99))` instead of `rescale_transform = RescaleIntensity((0, 1), (1, 99))`? Sure.

tests/transforms/augmentation/test_random_gamma.py
fepegar: Hmm let's see if I understand.

Sample is `[0.0 0.1, 0.5, 0.9, 1.0]`

**If log(gamma) = 100,**
then gamma = e^100 ~ 2.68^43 -> values in [0, 1) go to 0: transformed is `[0, 0, 0, 0, 1]`. That's why you compare sample == 1 with the result.

**If log(gamma) = -100,**
then gamma = e^-100 ~ 0 -> values in [0, 1) go to 1: transformed is `[1, 1, 1, 1, 1]`. Note that 0^0 = 1 in Python, so I think all values will go to 1.

1. I think you can also compare the rest of the values, making sure they are **almost** equal to the expected values.

torchio/transforms/augmentation/intensity/random_gamma.py
fepegar: I meant the `percentiles` kwarg.

torchio/transforms/augmentation/intensity/random_gamma.py
fepegar: I understand! I meant that we haven't used the `parameter_range` syntax in other transforms. For example, in `RandomAffine` we use `scales` and `degrees`, although ranges are also accepted.

tests/transforms/augmentation/test_random_gamma.py
GFabien: Yes that's it, OK, I will compare all values

torchio/transforms/augmentation/intensity/random_gamma.py
GFabien: Yes I don't know why I have written `log_gamma_range` instead of `log_gamma`, I will remove `_range`.

tests/transforms/augmentation/test_random_gamma.py
GFabien: In fact I am already looking at all the values because I compare the boolean maps to the whole tensors. What were you thinking when you were talking about comparing the rest of the values?

fepegar:
I have tried to make the docs a bit easier to understand (for me, at least). I also removed the recommendations to rescale, I think it should be the user's responsibility. Let me know if you disagree.

fepegar:
Added in `v0.17.37`. Thanks again, @GFabien!

PR #296, opened 2020-09-10
Faster implementation for UniformSampler

dvolgyes:
I found that the Uniform Sampler could be awfully slow for large volumes.
In my workflow, i have 512x512x ~800 large volumes, and i extract many patches  (e.g. 64) from each volume.
With these sizes, the extraction takes way slower time than the calculation of patch location.
This is because  UniformSampler uses the WeightedSampler, and WS calculates CDF every time,
this can be very time (and memory) consuming. This calculation grows linearly with the volume size,
and not with the patch size. 

However, generating the coordinates directly is cheap and simple, see the proposed implementation. 
I am not against keeping the get_probability_map method, if someone wants to use it for something else too,
(keeping the API similar for all methods, etc.), but i think it is unnecearry to use it for the calculations.
(But i am also not against removing it.)

Most of the code is trivial, one remark about the size: if the patch is as large as the volume, the valid range would be 0.
This is mostly fine, but randint is not inclusive, it will generate in 0..N-1 range, so we need to add +1, otherwise it would not be able to yield a patch as large as the volume.

### DIFF
--- a/torchio/data/sampler/uniform.py
+++ b/torchio/data/sampler/uniform.py
@@ -30,5 +30,5 @@ class UniformSampler(RandomSampler):
             raise RuntimeError(message)
 
         valid_range = sample.spatial_shape - self.patch_size
-        corners = np.random.randint(valid_range + 1)
+        corners = np.asarray([torch.randint(x+1,(1,)).item() for x in valid_range])
         yield self.extract_patch(sample, corners)
--- a/torchio/data/sampler/uniform.py
+++ b/torchio/data/sampler/uniform.py
@@ -1,10 +1,12 @@
 import torch
 from ...data.subject import Subject
 from ...torchio import TypePatchSize
-from .weighted import WeightedSampler
+from .sampler import RandomSampler
+from typing import Optional, Tuple, Generator
+import numpy as np
 
 
-class UniformSampler(WeightedSampler):
+class UniformSampler(RandomSampler):
     """Randomly extract patches from a volume with uniform probability.
 
     Args:
@@ -15,3 +17,18 @@ class UniformSampler(WeightedSampler):
 
     def get_probability_map(self, sample: Subject) -> torch.Tensor:
         return torch.ones(1, *sample.spatial_shape)
+
+    def __call__(self, sample: Subject) -> Generator[Subject, None, None]:
+
+        sample.check_consistent_spatial_shape()
+
+        if np.any(self.patch_size > sample.spatial_shape):
+            message = (
+                f'Patch size {tuple(self.patch_size)} cannot be'
+                f' larger than image size {tuple(sample.spatial_shape)}'
+            )
+            raise RuntimeError(message)
+
+        valid_range = sample.spatial_shape - self.patch_size
+        corners = np.random.randint(valid_range + 1)
+        yield self.extract_patch(sample, corners)
--- a/torchio/data/sampler/uniform.py
+++ b/torchio/data/sampler/uniform.py
@@ -19,7 +19,6 @@ class UniformSampler(RandomSampler):
         return torch.ones(1, *sample.spatial_shape)
 
     def __call__(self, sample: Subject) -> Generator[Subject, None, None]:
-
         sample.check_consistent_spatial_shape()
 
         if np.any(self.patch_size > sample.spatial_shape):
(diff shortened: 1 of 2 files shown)


torchio/data/sampler/uniform.py
```diff
@@ -15,3 +17,18 @@ def __init__(self, patch_size: TypePatchSize):
...
+
+        if np.any(self.patch_size > sample.spatial_shape):
+            message = (
+                f'Patch size {tuple(self.patch_size)} cannot be'
+                f' larger than image size {tuple(sample.spatial_shape)}'
+            )
+            raise RuntimeError(message)
+
+        valid_range = sample.spatial_shape - self.patch_size
+        corners = np.random.randint(valid_range + 1)
```
romainVala, line 33: Hi, 
you should avoid numpy random generation as it lead to bug witch torch dataloader
use torch.rand instead

dvolgyes:
I don't mind using torch, but could you provide any link supporting your claim? 
Sure, you need to set numpy seed for reproducibility, but this is needed anyway in any project since numpy is frequently used internally. But i never experiened any bug neither with single worker or multiple worker processes. So i would appreciate a pointer. 

But anyway, i will update it to torch.

romainVala:
sorry, I do not have link to support it, may be @fepegar .. ? 
it is just my own experience, but I may have missed something ...

but any way it is safer to have the same random generator throughout the code

dvolgyes:
Since the test (process_probabilities) makes no sense anymore (it is not used), i would just remove this test.
I am also leaning to remove the get_probability_map function, but i don't insist on it. Any opinion?

fepegar:
This is great, so much faster! I'll merge now.

**About NumPy vs PyTorch for RNGs**
I think this issue helped when I was in trouble: https://github.com/pytorch/pytorch/issues/5059#issuecomment-624518956

fepegar:
@all-contributors please add @dvolgyes for code.

fepegar:
Added in `v0.17.36`.

Thanks, @dvolgyes!

romainVala:
@dvolgyes : just curious about the application you are training ?
I test MRI cortical and subcortical segmentation and we had to sitch from UniformeSampler to WeightedSampler, if not it will be bad for the very small structure ...
Any chance to make WeightedSampler faster ?

dvolgyes:
I try to segment liver images: background, liver, tumor, but in a two phase scheme:
- background - liver on a smaller volume (downscaled),
- bounding box around the liver based on the downscaled
- patches from the bounding box for tumor segmentation
  (random for training, overlapping grid patches for the inference)
- some dice + focal loss, and some custom experiments, 
 but there is a weighting based on class frequency (except for dice, of course)

Brain segmentation: i am not familiar with the sizes, but i guess the above scheme could help.
Also, downscaling helps: you do not need exact locations, just approximately.
So what about:
- downscaling the original weight map with a factor of 4 by every dimension, this gives you 64x smaller weight map. 
- generate a position here, upscale it (x4 for every coordinate)
- add a random (0..3) offset in every direction
- make sure that the coordinates are valid (with the offset, you might move out)
- and take that patch

Of course, the downscaling takes time, and also cumsum, but only once. 
So you could store it as a precalculated volume, and load it when it is needed.
Processing the weight map is the expensive part now, not the patch extraction.
(You can cheat here a lot: e.g. if you use binary weights, then you can store 8 positions in a single byte,
so instead of using a float32 weight, you get 1 bit / pixel. With 64 downscaling, this is 512x smaller.
If you use precalculated cdf, then you need float32 (maybe float16), but still, the downsampling helps.
(Or you can combine: store 1 bit mask on disk, and downsample and calculate cumsum on the fly.)

Also frequently used technique: make a fixed list of possible patches, e.g. 1024, and add some local offset there.
This is almost as the downsampling, except that you should choose the blocks having equal probability,
so instead of searching, you can take it directly. This unequal block size (the volume is unequal because you fix the probabilities) makes the local offseting a bit more complicated, so probably not faster but more complicated than the downscaling.

I would guess that the current parallel queue and the improved weightedsampler should be fast enough together, depending on the volume size. It takes some extra time for the cdf, but shouldn't be too long.

fepegar:
I think we should discuss this in a new issue, but `np.argsort` is the most expensive operation in the weighted sampler. It takes about 15 seconds for Colin 2008, vs 1 second for a sampling map with half the size. Maybe we could add an option to use a downsampled (coarse) version of the given sampling map.

dvolgyes:
Well, downsampling should work, of course, but also, you just merged a version from me which removes argsort #308 ,
didn't this help?

fepegar:
It helped a little bit. I ran some tests and the run time for that line went from 6e-2 to 3e-5. Huge relative improvement, but tiny absolute one.

fepegar:
And you didn't remove argsort, but argmax.

PR #302, opened 2020-09-17
Multichannel label sampler

### LINKED ISSUE #291: Uniform sampling across classes
**🚀 Feature**

The proposed feature is the creation of a sampler that will take a label map and will have the same probability for each class of the label map to take a voxel of this class as the center of the sampled patch.

**Motivation**

After training some segmentation models, I've seen that if some classes were not present enough in the ground truth, performance dropped drastically. Sampling a bit more from these classes seems to solve the issue.

**Pitch**

I would like to have a sampler that would directly achieve that by just giving it the label map of interest. Maybe a list of labels we are interested in could be useful to sample only from the classes we want.

**Alternatives**

The current alternative is to use a `WeightedSampler` and provide it the probability map.

**Additional context**

The feature is rather simple, to achieve this behaviour, we just need to set the same value to every voxel and then divide the value of each voxel by the volume of its class (and finally normalize to have probabilities).

I think the feature is still generic enough to be relevant in TorchIO, do you agree? If you do I will provide a PR.

GFabien:
Fixes #291 
Update `LabelSampler` to change its behaviour if the input has more than 1 channel. In such a case, keys from the `label_probabilities` dictionary no longer represent label values but label channels instead.

### DIFF
--- a/torchio/data/sampler/label.py
+++ b/torchio/data/sampler/label.py
@@ -82,15 +82,25 @@ class LabelSampler(WeightedSampler):
             label_probabilities_dict: Dict[int, float],
             ) -> torch.Tensor:
         """Create probability map according to label map probabilities."""
+        multichannel = label_map.shape[0] > 1
         probability_map = torch.zeros_like(label_map)
         label_probs = torch.Tensor(list(label_probabilities_dict.values()))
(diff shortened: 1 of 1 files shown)


fepegar:
Nice. Thanks again, @GFabien!

PR #308, opened 2020-09-18
Faster WeightedSampler implementation

dvolgyes:
Np.sortedsearch performs binary search, which is logN steps, instead of
the arithmetics on the full array and then a linear search (argmax).


dvolgyes:
I meant np.searchsorted.

fepegar:
I was about to check this PR. Did you just push to the wrong branch?

dvolgyes:
Yes, but hopefully i undid it.

fepegar:
Added in `v0.17.40`. Thanks, @dvolgyes!

Issue #301, opened 2020-09-17
Parallelized Queue implementation


dvolgyes:
**🚀 Feature**
A processing queue implementation which has a limited capacity,
but immediately emits new data samples, and doesn't wait until the queue is full.

**Motivation**

Speed.

**Pitch**

Randomize in advance, but do not start the pre-loading until there is no empty spot in the queue.

**Alternatives**

Existing queue.

dvolgyes:
The primary question is about features. The parallel programming part is not easy, but also not too hard,
if the expected behavior is defined.

Basic assumptions:
- loading one volume and performing the augmentation takes a lot of time for each volume (30-60 seconds)
- an augmented volume consumes a lot of memory, so just a small portion of the dataset can be stored in memory (<10%)
- getting a patch out of a voume is / will be quick, especially with UniformSampler
- many patches (>2) or many many (>32) patches will be extracted. 
  If the volume can be reused, this is basically factor of N speedup.
- samples from a single volume are somewhat correlated, so it would be nice to mix samples from multiple volumes
- but also volumes could be / should be somewhat randomized.

Let's call the volume capacity V (how many volumes can be stored at the same time in memory),
and let's call the numer of patches P.

1, Most basic approach: take V (e.g. 4)  volumes and P (5) patches, and make (v,p) pairs, and randomize these:
(0,0), (3,2), (1,4), ...
When these run out, you can take the next batch, e.g. (5,0), (7,3), ...

The loader logic would be: load a volume when there is less then V volume in memory, and discard volume
when P patches were extracted. 

This is somewhat close to the current implementation, except the part of waiting for the queue to be filled, 
but raises a few question.
What happens if in the end less than V volumes are present? E.g. 4 volume long queue, 11volumes in tota.
- Should any previous volume be reused, if yes, which one? 
  The first ones (extra loading), or keeping from the previous batch (not discarding)?
- Could the final queue be (almost) twice as large? E.g. just merging the last two batches, 
  forming a super-batch with 7 volumes, and randomize the volumes here?

Or should be the last two batches merged, split into half, and make two shorter batch,
but both of them are at least V/2 size. (E.g. 4 and 2 would make 3+3).

2, A volume can be loaded multiple times, but also can be discarded. This allows more randomization,
   but it is not extra disk loading time, but the augmentations will be different for the same volume.

3, Some other approach.

My proposal:
A semi-intelligent parallel queue which takes a command stream. 
[...]

PR #311, opened 2020-09-21
Parallel queue

dvolgyes:
The code for #301. It is way uglier than it should be, but maybe it is easier to
understand this way. Tests, type hints, etc: i will fix it when we agreed on the details.
But now it should work.

In my examples the network processes a patch as fast as i can extract,
this will be a fundamenal limit. But if i put a time.sleep(1) into my training loop,
then i see that the queue preloads the volumes.

Recommendation: set the subject size to a small value, e.g. 3, the number of patches to a large number, like 64,
and put a time.sleep(1) into the training loop. Hopefully this will set conditions where you can study how it works.
(Assuming you have enough memory for that.)

### DIFF
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -15,3 +15,4 @@ tox
 twine
 watchdog
 wheel
+loguru
--- a/setup.py
+++ b/setup.py
@@ -21,6 +21,8 @@ requirements = [
     'torch>=1.1',
     'torchvision',
     'tqdm',
+    'unsync',
+    'loguru',
 ]
 
 setup(
@@ -58,6 +60,6 @@ setup(
     test_suite='tests',
     tests_require=[],
     url='https://github.com/fepegar/torchio',
-    version='0.17.38',
+    version='0.17.34',
     zip_safe=False,
 )
--- a/torchio/__init__.py
+++ b/torchio/__init__.py
@@ -2,7 +2,7 @@
 
 __author__ = """Fernando Perez-Garcia"""
 __email__ = 'fernando.perezgarcia.17@ucl.ac.uk'
-__version__ = '0.17.38'
+__version__ = '0.17.34'
 
 import os
 from . import utils
@@ -18,6 +18,7 @@ from .data import (
     ScalarImage,
     LabelMap,
     Queue,
+    ParallelQueue,
     Subject,
 )
 from . import datasets
--- a/torchio/data/__init__.py
+++ b/torchio/data/__init__.py
@@ -1,4 +1,5 @@
 from .queue import Queue
+from .parallel_queue import ParallelQueue
 from .subject import Subject
 from .dataset import SubjectsDataset, ImagesDataset
 from .image import Image, ScalarImage, LabelMap
--- a/torchio/data/parallel_queue.py
+++ b/torchio/data/parallel_queue.py
@@ -15,16 +15,29 @@ import multiprocessing as mp
 import torch
 import numpy as np
 from loguru import logger
+import sys
 
+class Counter(object):
+    def __init__(self, initval=0):
+        self.val = mp.Value('i', initval)
+        self.lock = mp.Lock()
+
+    def increment(self):
+        with self.lock:
+            self.val.value += 1
+
+    def value(self):
+        with self.lock:
+            return self.val.value
 
 # these are tools for the command stream, they will not be exported
 def shift_queue(queue, offset):
-    return [(C,i+offset) for (C,i) in queue]
+    return [(C,i+offset,L) for (C,i,L) in queue]
 
 def randomize_queue(queue):
-    num_subjects = sum(1 for C,idx in queue if C=='L')
+    num_subjects = sum(1 for C,idx,locks in queue if C=='L')
     new_indices = torch.randperm(num_subjects)
-    randomized_queue = [(C,new_indices[i].item()) for C,i in queue]
+    randomized_queue = [(C,new_indices[i].item(), tuple(new_indices[l].item() for l in locks)) for C,i,locks in queue]
     return randomized_queue
 
 def optimize_queue(queue):
@@ -37,7 +50,7 @@ def optimize_queue(queue):
     queue = queue[:]
     insert = 0
     for idx in range(1,len(queue)-1):
-        C,V = queue[idx]
+        C,V,L = queue[idx]
         if C != 'D':
             continue
         for pos in range(idx+1, len(queue)):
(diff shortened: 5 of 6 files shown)


sarthakpati:
Any idea when this will get merged?

fepegar:
@sarthakpati, according to the discussion in #301, it's not clear that this implementation would generally be preferred, and I believe it's now in a prototyping state. You can always check out this branch and use it, if you like. I'll convert this PR to a draft, so it's clear that it's not just waiting to be reviewed.

hsyang1222:
@fepegar, Would you consider providing a **parallel** queue? 
Despite torchio offers the huge convenience and efficiency, we have to spend a lot of time preprocessing images to 
preprocessing of the medical images. This preprocessing time is growing exponentially as datasets become higher quality, more diverse, and larger. The idea of learning on GPUs and parallelizing on CPUs used to feel like an experimental idea, but increasingly it seems to be a necessary methodology to solve the learning bottleneck.

I'm thinking of sticking with the current code interface, but offering this as an option in a parallel environment.

fepegar:
> @fepegar, Would you consider providing a **parallel** queue? Despite torchio offers the huge convenience and efficiency, we have to spend a lot of time preprocessing medical images. This preprocessing time is growing exponentially as datasets become higher quality, more diverse, and larger. The idea of learning on GPUs and parallelizing on CPUs used to feel like an experimental, but increasingly it seems to be a necessary methodology to solve the learning bottleneck.
> 
> I'm thinking of sticking with the current code interface, but offering this as an option in a parallel environment.

Unfortunately, I can't "provide" a parallel queue because I don't have the resources to work on it. I think nowadays the data loading tools available in modern versions of PyTorch would allow for a very nice implementation, but I really don't have the bandwidth. You're welcome to contribute!

hsyang1222:
@fepegar  I have some possible improvements to make queues more parallel. Can you tell me my understanding of queues is correct, and my improvement ideas are useful?

Understanding how queue works
(1) In queue, subjects are already loaded in parallel from main process via pytorch dataloader. If you set _num_workers_ >= 1, this dataloader prefetches data with _prefetch_ argument, so fetching data from the main storage (e.g. SSD) is already non-blocking. 
(2) The subjects fetched by the dataloader are moved to the main process when the queue calls _fill()_, after that transformed and cropped to the patch size.

I think this can be improved in two ways. 
(a) Handle the transform process inside the dataloader that prefetches the subject. This will reduce the amount of time blocked in creating patches because when _fill()_ is called, it will be passed subjects that have already been transformed.
(b) A typical transform consists of [the pretransform such as resampe, and the augmentation such as random noise]. The augmentation produces random results and must be executed every time, while the pretransform produces the same result every time so pretransform can be cached. In general, the resample time is long, so it's much better to save it to a file and then load it, even in a hard-disk environment. I want to automate this process: Split the transform into a pretransform and an augmentation, and cache the pretransform results.

fepegar:
> Handle the transform process inside the dataloader that prefetches the subject.

If I understood correctly, I think this is already the case. The data loader gets images from the dataset, where they are loaded and transformed in the subprocesses:

https://github.com/fepegar/torchio/blob/0933def2bc5eb61b8c3a15d59944cb09d22eceee/src/torchio/data/dataset.py#L82-L100

> A typical transform consists of [the pretransform such as resampe, and the augmentation such as random noise]. The augmentation produces random results and must be executed every time, while the pretransform produces the same result every time so pretransform can be cached. In general, the resample time is long, so it's much better to save it to a file and then load it, even in a hard-disk environment. I want to automate this process: Split the transform into a pretransform and an augmentation, and cache the pretransform results.

This is indeed interesting. I think it's related to MONAI's [`CacheDataset`](https://docs.monai.io/en/stable/data.html#cachedataset). I wonder what the gains would be. I guess if there is a `Resample` transform used for preprocessing, caching could help quite a lot.

dvolgyes:
Hi,

I used to write the patch long time ago, and as @fepegar said: it was not clear if it would improve speed. 
Unfortunately, I have no time to fix it, but if someone picks it up, here are my past year experiences:

- Parallel data loader is nice, but quite tricky. Usually the transforms make it complicated, and CPU based transforms are slow. And you can't use GPU based transforms in worker processes.
- The disk access is a huge bottleneck, it would be preferable to use some hiearchical structure which can read subsets properly and not in DICOM format. Remote sensing use tiled datasets, that is actually reasonably implementable in arbitrary dimension. (You select a griding, e.g. 64**3, that will be the underlying storage, and a slicing operator intelligently reads the blocks. E.g.  x[58:66, ...] would read two blocks (along first dim) but not the rest.) Current medical format usually read the whole volume into memory first, which is quite expensive. So pre-processing a volume into a tiled volume is probably the most fruitful approach. I would probably try tensorstore first before I would try to implement anything on my own. 
https://google.github.io/tensorstore/

Static transforms could be applied before the dataset, and nonstatic afterwards, something like

```
noncached_dataset = Dataset(..., transform=cached_transforms)
dataset = CachedDataset(noncached_dataset, cache_dir=..., expiration_time=, transforms=...)
```

where the __getitem___(self, idx): would be something like this:
```
def __getitem___(self, idx):
    if idx in self.cache:
        data = self.cache[idx]
    else:
        data = self.original_dataset[idx]  # cached transform is passed in original dataset
        self.cache[idx] = data
    if self.transform:
        data = self.transform(data)
    return data
```

So what do you need? First, even if you have N worker processes, only one should create the cache.
That should be even shared between multiple trainings. (imagine a server with 4 gpus, and you have parallel trainings)

However, this design has a serious limitation, that is CUDA access. Workers don't have CUDA access.
So probably it is a better way for going no worker processes, and give cuda access to the dataset object.
In this case, the dataset could make subprocesses, but probably shouldn't. 
After years of maturing the idea, I would probably go for an async data storage where you can send many async calls, and the storage backend handles both parallelization, caching, ordering, etc., (see tensorstore)
The most important part here is the locking mechanism: that is error-prone to implement, and better to leave it to the storage than using multiprocessing locks. 

In remote sensing, there is another thing: coordinate systems. While this is not directly applicable to medical imaging, the different modalities exhibit similar resampling challenges. 
I would recommend taking a look at https://torchgeo.readthedocs.io/en/stable/
Both the coordinate system handling and the spatial samplers are interesting (random, grid, prechipped, etc.)

So summarizing: I would guess the disk format is the most significant bottleneck, and the only reasonable way, I think, it to read the pixel data from a preprocessed file, and as a first attempt I would go for tensorstore. 
As of structure: I would go for async variant of tensorstore, and not using any subprocesses. This makes it possible to run CUDA based transforms asynchronously from the model. (just careful with free memory)

Ugly but not completely out of question: most medical datasets are small: make an in memory cache, load everything in the beginning of the training, and make subsamples from these. (unfortunately, we rarely get more than 100G data, and 100G in a server is not too much).

Monai & co: I would assume that everyone has the same problem, so sooner or later someone will solve it. But most frameworks have relatively clear cut at the dataset level, so it is not unreasonable to combine datasets/loaders/samplers with the other frameworks transforms, models, losses, etc.

hsyang1222:
Thank you, @fepegar and @dvolgyes , for your kind responses. I mentioned improving the prefetch of __queue__ (a) and caching the pretransform (b). I'll talk about those two first, and then talk about a broader speedup (c).

(a) Improving prefetch on __queue__
I think fepegar is right about queue. The dataloader currently uses prefetch to fetch data asynchronously, but I still think it could be improved.
I'll create a separate issue to make a pull request for this.

(b) Caching pretransforms
Like fepegar and dvolgyes said, I think a CachedDataset could be helpful. I didn't know about CachedDataset when I started talking about this, so I hadn't considered it up front. Currently I simply (b-1) load the dataset from disk and create a __ScalarImage(tensor= data from disk, affine = affine from disk)__ to store the data in memory, (b-2) perform the pretransform, and (b-3) only handle the augmentation when training. The speedup was about 5x per iteration.
 However, this method was difficult to use in all common situations: memory management was difficult, and even a small increase in data moved data to swap memory, which was sometimes slower than just fetching data from disk.
 Now that I know about CachedDataset, I'm sure there's research and libraries out there for this situation, and I'll be looking into it.

(c) Using CUDA for transforms
 I used kornia (https://github.com/kornia/kornia) to perform augmentation via GPU. However, some features (elastic deformation, pretransform) are not supported and there are differences in the settings for each feature. The use of GPUs for augmentation does not seem to be fully consensual. For example, in my case, when training a large model, I use torchio transform as it is due to lack of GPU memory.

Thanks again for your answers. I think there is still a lot of room for improvement in medical image pretransforms. If I can think of more, I'll comment again.

romainVala:
Hi there,

Just my feedback about "pre-transform" because I do not think that this is the major bottleneck. Of course it may depend on the exact choice of transform (depending on your application). In my case I do use a RandomAffine, and a RandomElastics  ...  even worst I also find very useful to use  RandomMotion ! So each on those transforms takes a quite a lot of time. I would love to see them implemented on GPU, (should it be a developing objective of torchio ???) 
Anyway for now we need to deal with cpu transform ... For an efficient training I needed more than 32 cpu core (and a num_worker above 12) so that the gpu does not wait too long. But increasing num_worker does also goes with increasing (cpu) memory

Unfortunately the gpu node, I can access, do not have enough cpu memory and reducing the num_worker (to 6) lead to training time just too long (a factor 5 to 10 or more compare with almost no augmentation)

I end up  doing the augmentation on a cpu cluster and saving each iteration on disk. This is a quite insane solution since I do need a few Terra of disk, to save my few hundred thousand augmented examples, but it is effective and I greatly speed up the training

fepegar:
Hi all. Thank you for your contributions and discussions. I'm going to close this PR as it's been stale for years. Happy to reopen if needed, or we can create a discussion somewhere else.

Issue #312, opened 2020-09-21
Faster implementation for WeightedSampler


fepegar:
Discussion about how to make the `WeightedSampler` faster. It stems from a discussion with @dvolgyes and @romainVala in #296.

Related to #308.

@dvolgyes:

> WS calculates CDF every time,
this can be very time (and memory) consuming. This calculation grows linearly with the volume size

The bottleneck seems to be in `np.argsort`:

https://github.com/fepegar/torchio/blob/c3c5e77d2cddfdff03ad00453a76d3f76c360061/torchio/data/sampler/weighted.py#L188

I think a good solution, proposed by @dvolgyes and myself, is downsampling the sampling map for faster computation. The downsampling factor can be passed by the user as a kwarg.

dvolgyes:
Well, as you said, i didn't remove argsort, but actually i wanted.
I struggle to see why is it needed.
I think it is not needed at all.
You do not needd to order probabilities.
Cumsum will make an increasing array (cdf), you take a sample from this with searchsorted,
and convert it back to coordinates with unravel.
Sorting is both expensive and unnecessary.
(And you can even introduce bugs with it, and i have to think, because at this moment i don't see
if the current implementation is correct or not.)
Anyway, so i meant to remove argsort and argmax at the same time, and that would make it quite fast. :)
(Regardless of downsampling.)

fepegar:
The implementation is inherited from NiftyNet's, I understood it at the time and reimplemented it here. It's great that you have noticed that some operations are not needed. I'll check that PR.

dvolgyes:
Not counting probable some style issues, the code should work. 
Check it out if it is a) fast enough, b) works as intended. :)

fepegar:
Improvement on CDF computations added by @dvolgyes on `v0.17.41`. Closing this for now. Will reopen if we want to add the downsampling feature.

@OeslleLucena, lmk if you're interested in working on that!

PR #313, opened 2020-09-21
Even faster WeightedSampler

dvolgyes:
See #312.
I blind-coded it, so let's see if i was right, or the CI fails. :)
(I didn't have a weighted sampler set up, but i will do it soon.)


fepegar:
I ran this code

```python
import numpy as np
import torchio as tio
from tqdm import tqdm

probability_map = np.array(
    (0,0,1,1,5,2,1,1,0,
    2,2,2,2,2,2,2,2,2),
).reshape(1, 1, 1, -1)

im = np.ones_like(probability_map)
hist = np.zeros(probability_map.shape[-1])

subject = tio.Subject(
    im=tio.ScalarImage(tensor=im),
    pm=tio.Image(tensor=probability_map, type=tio.SAMPLING_MAP)
)

sampler = tio.data.WeightedSampler(1, 'pm')

indices = []
n = 10000
for patch in tqdm(sampler(subject, n), total=n):
    index = patch['index_ini'][-1]
    hist[index] += 1

for a, b in zip(probability_map.ravel().tolist(), hist.tolist()):
    print(f'{a}\t{int(b)}')
```

on this branch and the results are as expected:

```
0	0
0	0
1	329
1	356
5	1674
2	699
1	330
1	346
0	0
2	652
2	704
2	693
2	702
2	697
2	720
2	694
2	691
2	713
```
so this looks good to me. I'm not sure how to include that verification into good unit tests, though.

fepegar:
Wait, I didn't switch branches properly.

dvolgyes:
Unit test: maybe giving a few extreme values, e.g. 32x32x32 volume, and setting weights
for 3 random locations, e.g. 7,12,21 and 11,13,15 and 23,8,16 to a large value, e.g. a billion,
and checking that the returned position is one of the three. (Be careful with centers vs. corners.)
And if it fails every blue moon (1 in a billion), then we re-run the tests? (if the other values are not zeros.
if they are, then this set should  always be fine.)

fepegar:
I'm timing the code in `__call__`, before and after using Colin27 2008.

**Before**
8.640887022018433
6.1830809116363525
5.196950912475586
6.345213890075684
13.973642826080322
12.116162061691284
11.626728773117065
10.925238132476807
11.24952483177185
7.559868812561035
10.074389219284058
5.8907997608184814
5.609090089797974
5.424769878387451
5.8451972007751465
5.591778039932251
5.30805516242981
9.0135018825531
6.34114933013916
5.31603479385376

**After this PR**
4.581050872802734
3.905184030532837
2.3259856700897217
2.8281338214874268
2.467862129211426
2.9173831939697266
2.566795825958252
2.7686421871185303
2.368414878845215
3.2915608882904053
3.31826114654541
2.2530770301818848
2.4584991931915283
2.4432599544525146
2.872347831726074
3.2814629077911377
2.3743410110473633
2.445211887359619
2.5871827602386475
2.806492328643799

So I guess ~5x faster!

> Unit test: maybe giving a few extreme values, e.g. 32x32x32 volume, and setting weights
for 3 random locations, e.g. 7,12,21 and 11,13,15 and 23,8,16 to a large value, e.g. a billion,
and checking that the returned position is one of the three. (Be careful with centers vs. corners.)
And if it fails every blue moon (1 in a billion), then we re-run the tests? (if the other values are not zeros.
if they are, then this set should always be fine.)

I think that's similar to the current one, except we only set one voxel to 1. And I know it's really improbable to fail, but ideally we want deterministically successful tests...

dvolgyes:
I guess the bottlenecks now are: cumsum (and disk IO for the raw data).
It is a bit strange that e.g. the 5th went down to 2.5 (from 14), while the first went down 4.6 to (from 8.6).
Do you measure the __call__ of the sampler, or some other method's __call__?
Maybe there are still background processes in the beginning, which reduces CPU capacity in the beginning,
so maybe a line-by-line profiling (with single worker process, etc.) would highlight what is left.

But i strongly believe it is the cumsum, and this one cannot be really worked-around without downsampling.

dvolgyes:
Anyway, i also believe that the parallel queue implementation should help at least 2-3 seconds/subject, so i would say the weighted sampler is now fast enough. The subject loaders and/or the transformations are probably the next major bottlenecks.

fepegar:
Yeah, my benchmarking wasn't very formal. It was just to get a rough idea of the order of magnitude.

Alright, merging this and closing the corresponding issue for now. We can always add the downsampling later on.

PR #314, opened 2020-09-21
Fix LabelSampler and add test

GFabien:
I must have made a mistake when creating the PR the other day as there was an additional line creating an error in case of multichannel label map. I removed it and added a test to make sure it's fine now.
Sorry about that.


fepegar:
Great. Thanks, @GFabien!

Issue #315, opened 2020-09-21
Add pre-commit hooks for more standardized development


dvolgyes:
**🚀 Feature**
Add pre-commit hooks through the "pre-commit" framework.
[https://pre-commit.com/](https://pre-commit.com/)

**Motivation**

Basically it is a very nice way to reduce code issues: formatting, old code style, mutable parameters, etc.
There are literally hundreds of checks out there, and the installation is trivial.
Every developer on their computer needs two extra lines:
pip install pre-commit
(enter the torchio git directory)
pre-commit install

The code styles, checks, etc., are defined in the .pre-commit-config.yaml, so it is project-wise standardized.

**Pitch**

It is good.

**Alternatives**
Like without CI: good-old non-testing.

**Additional context**

An example what i use elsewhere, but it shows how to configure the project.
I will open a PR with a simple one which doesn't use auto-formating, only other issues.
Autopep8 or black or others could automatically enforce style, but you need to define your own,
and then we can put it into the config file. 

```
repos:
-   repo: local
    hooks:
    -   id: exclude-unwanted
        name: exclude unwanted files, e.g. training data
        entry: exclude unwanted files, e.g. training data
        language: fail
        files: '.*png|.*xlsx'
-   repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v2.3.0
    hooks:
    -   id: trailing-whitespace
    -   id: check-added-large-files
    -   id: check-case-conflict
    -   id: check-ast
    -   id: check-executables-have-shebangs
    -   id: check-json
    -   id: check-merge-conflict
    -   id: check-symlinks
    -   id: check-toml
    -   id: check-xml
    -   id: check-vcs-permalinks
    -   id: check-xml
    -   id: detect-private-key
    -   id: end-of-file-fixer
    -   id: pretty-format-json
    -   id: requirements-txt-fixer
    -   id: sort-simple-yaml
    -   id: double-quote-string-fixer
    -   id: mixed-line-ending
    -   id: debug-statements
    -   id: pretty-format-json
        args: ['--autofix']
    -   id: flake8
        additional_dependencies: [
            flake8-bandit,
            flake8-blind-except,
            flake8-breakpoint,
            flake8-broken-line,
            flake8-bugbear,
            flake8-builtins,
            flake8-class-newline,
            flake8-comprehensions,
            flake8-debugger,
            flake8-eradicate,
            flake8-logging-format,
            flake8-pep3101,
            flake8-polyfill,
[...]

PR #316, opened 2020-09-21
Add pre-commit support

dvolgyes:
Add pre-commit support for development, see #315.
The actual checks might need customization, it reflects mostly my preference,
and i already commented out some which are whining a lot.
(I would use them, but it is not my project.)

Also i have an auto code formating hooks (autopep8), but some people prefer
black, or yapf, or else. 

If you check out the config, you will see how to disable invidiual reports.

Caching: first run takes time, the rest is fast. (Try run it twice in a row.)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,7 +4,6 @@ repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v3.2.0
     hooks:
-      - id: check-added-large-file     # Prevent giant files from being committed
       - id: check-ast                  # Simply check whether the files parse as valid python
       - id: check-case-conflict        # Check for files that would conflict in case-insensitive filesystems
       - id: check-builtin-literals     # Require literal syntax when initializing empty or zero Python builtin types
@@ -17,6 +16,8 @@ repos:
       - id: end-of-file-fixer          # Ensures that a file is either empty, or ends with one newline
       - id: mixed-line-ending          # Replaces or checks mixed line ending
       - id: trailing-whitespace        # This hook trims trailing whitespace
+      - id: file-contents-sorter       # Sort the lines in specified files
+        files: ^requirements.*\.txt$
   - repo: https://gitlab.com/pycqa/flake8
     rev: 3.8.3
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,72 @@
+repos:
+-   repo: local
+    hooks:
+    -   id: exclude-unwanted
+        name: exclude unwanted files, e.g. training data
+        entry: exclude unwanted files, e.g. training data
+        language: fail
+        files: '.*xlsx|.*dcm|.*pyc'
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v3.2.0
+    hooks:
+    -   id: trailing-whitespace
+    -   id: check-added-large-files
+    -   id: check-case-conflict
+    -   id: check-ast
+    -   id: check-executables-have-shebangs
+    -   id: check-json
+    -   id: check-merge-conflict
+    -   id: check-symlinks
+    -   id: check-toml
+    -   id: check-xml
+    -   id: check-vcs-permalinks
+    -   id: check-xml
+    -   id: detect-private-key
+    -   id: pretty-format-json
+    -   id: requirements-txt-fixer
+    -   id: sort-simple-yaml
+    -   id: mixed-line-ending
+    -   id: debug-statements
+    -   id: pretty-format-json
+    -   id: end-of-file-fixer
+        args: ['--autofix']
+-   repo: https://gitlab.com/pycqa/flake8
+    rev: 3.8.3
+    hooks:
+    -   id: flake8
+        additional_dependencies: [
+#            flake8-bandit,
+            flake8-blind-except,
+            flake8-breakpoint,
+            flake8-broken-line,
+            flake8-bugbear,
+            flake8-builtins,
+            flake8-class-newline,
+            flake8-comprehensions,
+#            flake8-debugger,
+#            flake8-eradicate,
+#            flake8-logging-format,
+#            flake8-pep3101,
+#            flake8-polyfill,
+#            flake8-quotes,
+#            flake8-string-format,
+            flake8-type-annotations,
+#            flake8-use-fstring,
+#            flake8-fixme,
+            flake8-markdown,
+            flake8-2020,
+#            flake8-print,
+#            pep8-naming
+             ]
+        args: ['--max-line-length=160', '--ignore=E121,E123,E125,E501,F401,F403,W503']
+
+-   repo: https://github.com/pre-commit/pygrep-hooks
+    rev: v1.6.0  # Use the ref you want to point at
+    hooks:
+    -   id: python-check-blanket-noqa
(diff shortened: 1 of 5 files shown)


fepegar:
I have customized it a bit and fixed many things pointed out by some hooks I uncommented.

I love `pre-commit`!

I have added instructions to run `pre-commit` install in the `CONTRIBUTING` docs. I'll also create a pull request template. Does that sound right?

dvolgyes:
I am glad that you like it, i love it too, hence the recommendation. :)
(I also recommend to add it to cookiecutter templates, so in next project you don't start from scratch.
What i haven't done, but i consider is having a meta-pre-commit hooks:
having a hook to check that my current hook is the latest from a given git repository. 
I have many projects, and keeping in mind which one was updated is kind of annoying,
but in my own projects i can enforce the same hook in every project. I will probably write a quick hook
for this tonight. (Same could work for .gitignore files, and a few others.))

I like pre-commit and propose it everywhere because in different project the contributions guides are different,
and it is a nightmare to remember them all, or read the contribution guide before every commit. 

Template and contributing docs: i am not sure, but it seems you might have forgotten to add it to the git commits.
Anyway: yeah, in general this should be the way: giving pointer in the contrib docs. 
Two remarks: besides pre-commit, pipx is a great tool. Similarly to pre-commit it creates virtual environment for command line tools, and it keeps them separate. I use pipx outside of conda (using system python), so when i switch conda environment, pipx still find my pre-commit. If you haven't use pipx yet, try it, you will love it too. 

Long term: pre-commit.ci is in progress and the idea, if i understood it correctly, that they will provide a pre-commit bot, so when there is a pull request it can not only check conformance, but also e.g., reformat code, and all these before or in parallel with CI/CD. So if the developer doesn't install pre-commit, on the pull request it would still run. But it is not yet ready.
See / hear it here: [https://talkpython.fm/episodes/show/282/pre-commit-framework](https://talkpython.fm/episodes/show/282/pre-commit-framework)

Anyway: i think the ```CONTRIBUTING``` wasn't committed, otherwise it looks good. :)

dvolgyes:
You might want to check this out too: how to keep pre-commit updated in git.
https://browniebroke.com/blog/gh-action-pre-commit-autoupdate/

The point is: sometimes the tools are updated, especially the flake8 based bug hunting tools, like bugbear, 
so using the latest version from a given hook is a good idea. And github can run scheduled actions for this.

fepegar:
I got overexcited and changed `CONTRIBUTING` on `master` instead of here. I will add the PR template soon.

I'll take a look at `pre-commit.ci`. But I guess it's not very convenient to have to wait until the PR to know if the code is ok for the repo. However I agree that it's safer than assuming that everyone will have installed `pre-commit`, so I suppose it wouldn't hurt.

Let me know if you write that meta-hook! I'll also check out the GH action. I still don't know anything about them.

I'll take a look at `pipx`. After a quick look, I'm not sure what it can do that `conda` cannot.

Thanks for your contribution and your advice!

dvolgyes:
pipx: this is easy.
Let's assume these:
```
conda create env1
conda install ... pre-commit
pre-commit install 
conda create env2
# switch to env2
git commit ...
```
In this case pre-commit, obviously, might have some issues. It might still find the original code,
because it kept a pointer into the conda environment, but should it? 
I mean, what if you remove this environment?
Or do you install pre-commit for every conda environment every time? What if you forget?
If you install pipx, it will make a link outside of conda, install the command into a closed, independent 
virtual environment which has nothing else but the given command installed (and dependencies, obviously)
Also, it can do these:
```
pipx list
pipx upgrade-all
pipx uninstall-all
pipx reinstall-all
```
Especially, the upgrade-all is super comfy. 
Basically, i use this way:
```
system-python -m pip install pipx 
system-python -m pipx ensurepath
pipx install pre-commit
pipx install cookiecutter
pipx inject cookiecutter    other packages i need
...
```

Basically, it gives a separate environment which is based on pip, not conda, 
and if you switch conda environment, the executables still remain accessible,
but at the same time everything is isolated.
Even more: you can auto-update packages. In conda you can only auto-update _conda_ packages, but not pip installed.

The only trick is the first pipx installation. It has its own magic to find stable python,
but i like to avoid confusion with conda, so i just call the system python (/usr/bin/python),
and once again for ensuring path, but after that it should be able to manage itself.

fepegar:
I'll take a closer look and give it a try. Thanks again!

dvolgyes:
It is a bit ugly, it was written after midnight, but you can try this: 
https://github.com/dvolgyes/pre-commit-sync
It works like this:
```
repos:
-   repo: https://github.com/dvolgyes/pre-commit-sync
    rev: "v0.1.1"
    hooks:
    -   id: pre-commit-sync
        args: ['--sync=.pre-commit-config.yaml:https://raw.githubusercontent.com/fepegar/torchio/master/.pre-commit-config.yaml']
```
In the master repository, e.g. fepegar/torchio it needs to reference itself, in the derived repos, you copy the file once,
and it will keep itself up-to-date. You can also add it to github actions, so it will create pull requests without anybody actively committing. But i guess this is overkill.

It is a bit tricky, you need to think on it a bit, but basically the hook takes an URL and saves it into a file, that is it.
But with a little self-referencing trick, this is enough to keep multiple repositories in sync. 
If it is used for anything else, e.g. gitignore or license, then it becomes trivial.

fepegar:
I modified the yaml with your hook, and installed and ran pre-commit. It deleted itself!

```
pre-commit installed at .git/hooks/pre-commit
Check python ast.........................................................Passed
Check for case conflicts.................................................Passed
Check builtin type constructor use.......................................Passed
Check docstring is first.................................................Passed
Check for merge conflicts................................................Passed
Check vcs permalinks.....................................................Passed
Debug Statements (Python)................................................Passed
Detect Private Key.......................................................Passed
Fix double quoted strings................................................Passed
Fix End of Files.........................................................Passed
Mixed line ending........................................................Passed
Trim Trailing Whitespace.................................................Passed
File Contents Sorter.....................................................Passed
flake8...................................................................Passed
check blanket noqa.......................................................Passed
use logger.warning(......................................................Passed
type annotations not comments............................................Passed
rst ``code`` is two backticks............................................Passed
rst ``inline code`` next to normal text..................................Passed
pyupgrade................................................................Passed
pre-commit-sync..........................................................Failed
- hook id: pre-commit-sync
- exit code: 1
- files were modified by this hook

File has been updated: .pre-commit-config.yaml
  (source: https://raw.githubusercontent.com/fepegar/torchio/master/.pre-commit-config.yaml)
File has been updated: .pre-commit-config.yaml
  (source: https://raw.githubusercontent.com/fepegar/torchio/master/.pre-commit-config.yaml)
File has been updated: .pre-commit-config.yaml
  (source: https://raw.githubusercontent.com/fepegar/torchio/master/.pre-commit-config.yaml)
```

What do you suggest? Committing this without verifying?

dvolgyes:
Hi,

Conceptually it is a bit a snake eating its own tail. 
Basically the hook doesn't do anything else, just takes a file and downloads it to somewhere. 
So if it is references itself, it will overwrite itself. 
However, this is quite inconvenient, but 2am in the morning this seemed to be a good idea.

Well, since then i created  a new rule: if the file is in git and locally modified, it will not update itself.
This way you can just commit the changes, no need for '--no-verify' anymore.

Basically, all you need to do is to update the hook to v0.2.0, and then it should work,
unless i forgot another corner case. :)

PR #319, opened 2020-09-23
Update rescale.py

neuronflow:
fix typo: nnU-Net uses 0.005 and 0.995 percentiles, see:
https://github.com/MIC-DKFZ/nnUNet/blob/4d8aa747b288e4297236eb385ea325256634372c/nnunet/preprocessing/preprocessing.py#L278

if percentiles should be provided without the leading zero then it needs to say: 0.5 to 99.5 percentiles, however this would be a very unusual notation.

### DIFF
--- a/torchio/transforms/preprocessing/intensity/rescale.py
+++ b/torchio/transforms/preprocessing/intensity/rescale.py
@@ -20,7 +20,7 @@ class RescaleIntensity(NormalizationTransform):
         percentiles: Percentile values of the input image that will be mapped
             to :math:`(n_{min}, n_{max})`. They can be used for contrast
             stretching, as in `this scikit-image example`_. For example,
-            Isensee et al. use ``(0.05, 99.5)`` in their `nn-UNet paper`_.
+            Isensee et al. use ``(0.005, 0.995)`` in their `nn-UNet paper`_.
             If only one value :math:`d` is provided,
             :math:`(n_{min}, n_{max}) = (0, d)`.
         masking_method: See
--- a/torchio/transforms/preprocessing/intensity/rescale.py
+++ b/torchio/transforms/preprocessing/intensity/rescale.py
@@ -20,7 +20,7 @@ class RescaleIntensity(NormalizationTransform):
         percentiles: Percentile values of the input image that will be mapped
             to :math:`(n_{min}, n_{max})`. They can be used for contrast
             stretching, as in `this scikit-image example`_. For example,
-            Isensee et al. use ``(0.005, 0.995)`` in their `nn-UNet paper`_.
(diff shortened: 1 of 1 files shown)


fepegar:
I'm not sure I follow.

In that repo I see the following code:

```python
                lower_bound = self.intensityproperties[c]['percentile_00_5']
                upper_bound = self.intensityproperties[c]['percentile_99_5']
```

Which looks like `0.5` and `99.5`. Moreover, as far as I know, percentiles go from 0 to 100. So we are using the 0.5th and 99.th percentiles to compute the corresponding percentile values: https://en.wikipedia.org/wiki/Percentile#Worked_examples_of_the_nearest-rank_method

I believe this is also NumPy's notation: https://numpy.org/doc/stable/reference/generated/numpy.percentile.html

Your suggestion is the one used in NumPy's quantile function: https://numpy.org/doc/stable/reference/generated/numpy.quantile.html#numpy.quantile

neuronflow:
Yes, just checked the syntax for np.percentile and was about to change my pull request. Indeed 0.5 and 99.5 seem correct :)

My background is more in R programming and there percentiles are usually denoted between 0 and 1 as they are PERCENTiles :7

neuronflow:
I currently how one would mirror the nnUnet pipeline in tio? What would be right parameters especially for `out_min_max`?
The rescaling from nnU-Net seems to be applied per channel but unlike Normalization Rescaling has no `channel_wise` flag?
Then nnU-Net brings in normalization so I guess rescaling and normalization from tio should be combined?

fepegar:
Are there any code changes in this PR?

> My background is more in R programming and there percentiles are usually denoted between 0 and 1 as they are PERCENTiles :7

Percents go from 0 to 100 :)

> I currently how one would mirror the nnUnet pipeline in tio? What would be right parameters especially for out_min_max?
The rescaling from nnU-Net seems to be applied per channel but unlike Normalization Rescaling has no channel_wise flag?
Then nnU-Net brings in normalization so I guess rescaling and normalization from tio should be combined?

Can you please open a new issue to discuss the nnU-Net stuff? I think for `out_min_max` you need the tuple I wrote in the docstring. We can discuss the rest in the issue.

neuronflow:
> Are there any code changes in this PR?

No, only correcting the comment.

> Percents go from 0 to 100 :)

From my limited understanding e.g. 42 percent  is short for the fraction 42/100 (42 per <cent> derived from Latin for  hundred) and this equals to 0.42? :7

> Can you please open a new issue to discuss the nnU-Net stuff? I think for `out_min_max` you need the tuple I wrote in the docstring. We can discuss the rest in the issue.

Hope I find time for that tomorrow night? :)

fepegar:
@all-contributors please add @neuronflow for docs

PR #328, opened 2020-10-07
Added support for HDF5 Dataset as Image including lazy loading of patches

soumickmj:
Will be able to use HDF5 datasets instead of path or tensor for creation of Image. The dataset can either be in (CxHxWxD) or even (HxWxD) format. If its the later, then the data will be expended for channel dim.
With the lazypatch parameter, will be able to enable or disable lazy loading of patches (only the patched portion of the volume will be loaded in the memory from the HDF5 Dataset).
 
### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -109,3 +109,4 @@ ENV/
 test_scripts/
 
 *DS_Store
+tester.py
--- a/setup.py
+++ b/setup.py
@@ -21,6 +21,7 @@ requirements = [
     'torch>=1.1',
     'torchvision',
     'tqdm',
+    'h5py'
 ]
 
 setup(
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -4,6 +4,7 @@ import random
 import tempfile
 import unittest
 from pathlib import Path
+import h5py
 
 import torch
 import numpy as np
@@ -168,3 +169,34 @@ class TorchioTestCase(unittest.TestCase):
 
     def assertTensorAlmostEqual(self, *args, **kwargs):  # noqa: N802
         assert_array_almost_equal(*args, **kwargs)
+
+    def get_h5DS_path(
+            self,
+            stem,
+            binary=False,
+            shape=(10, 20, 30),
+            spacing=(1, 1, 1),
+            components=1,
+            add_nans=False,
+            force_binary_foreground=True,
+            no_channel_dim=False
+            ):
+        shape = (*shape, 1) if len(shape) == 2 else shape
+        if no_channel_dim:
+            data = np.random.rand(*shape)
+        else:
+            data = np.random.rand(components, *shape)
+        if binary:
+            data = (data > 0.5).astype(np.uint8)
+            if not data.sum() and force_binary_foreground:
+                data[..., 0] = 1
+        if add_nans:
+            data[:] = np.nan
+        affine = np.diag((*spacing, 1))
+        suffix=".h5"
+        path = self.dir / f'{stem}{suffix}'
+        if np.random.rand() > 0.5:
+            path = str(path)
+        with h5py.File(path, "w") as f:
+            f.create_dataset("data", data=data)
+        return path
--- a/torchio/data/image.py
+++ b/torchio/data/image.py
@@ -61,6 +61,11 @@ class Image(dict):
         tensor: If :py:attr:`path` is not given, :attr:`tensor` must be a 4D
             :py:class:`torch.Tensor` or NumPy array with dimensions
             :math:`(C, W, H, D)`.
+        h5DS: If both path and tensor aren't supplied, then an HDF5 Dataset has
+            to be supplied. 
+        lazypatch: Only applicable with h5DS. If ``True``, the h5DS won't be 
+            pre-loaded and lazy-loading will be performed while loading part 
+            of the volume as patch. If ``False``, the whole h5DS will be preloaded. 
         affine: If :attr:`path` is not given, :attr:`affine` must be a
             :math:`4 \times 4` NumPy array. If ``None``, :attr:`affine` is an
             identity matrix.
@@ -97,6 +102,8 @@ class Image(dict):
             path: Union[TypePath, Sequence[TypePath], None] = None,
             type: str = None,
             tensor: Optional[TypeData] = None,
+            h5DS: Optional[TypeData] = None,
+            lazypatch: bool = False,
             affine: Optional[TypeData] = None,
             check_nans: bool = False,  # removed by ITK by default
             channels_last: bool = False,
@@ -113,8 +120,8 @@ class Image(dict):
             )
             type = INTENSITY
 
-        if path is None and tensor is None:
-            raise ValueError('A value for path or tensor must be given')
+        if path is None and tensor is None and h5DS is None:
+            raise ValueError('A value for path or tensor or h5DS must be given')
         self._loaded = False
 
         tensor = self.parse_tensor(tensor)
@@ -127,6 +134,14 @@ class Image(dict):
             if key in kwargs:
                 message = f'Key "{key}" is reserved. Use a different one'
                 raise ValueError(message)
+
+        if h5DS is not None: #as no other affine is currently supplied inside the dataset
+            self[AFFINE] = affine
+
+        self.h5DS = h5DS
+        self.lazypatch = lazypatch
+        if h5DS and not lazypatch:
+            self.load()
 
         super().__init__(**kwargs)
         self.path = self._parse_path(path)
@@ -194,7 +209,10 @@ class Image(dict):
 
     @property
     def spatial_shape(self) -> TypeTripletInt:
-        return self.shape[1:]
+        if len(self.shape) == 3:
+            return self.shape
+        else:
+            return self.shape[1:]
 
     def check_is_2d(self):
         if not self.is_2d():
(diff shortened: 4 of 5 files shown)


soumickmj:
@fepegar Here is the new pull request with fresh single commit

tests/utils.py
```diff
@@ -168,3 +169,34 @@ def assertTensorEqual(self, *args, **kwargs):  # noqa: N802
...
+        if binary:
+            data = (data > 0.5).astype(np.uint8)
+            if not data.sum() and force_binary_foreground:
+                data[..., 0] = 1
+        if add_nans:
+            data[:] = np.nan
+        affine = np.diag((*spacing, 1))
+        suffix=".h5"
+        path = self.dir / f'{stem}{suffix}'
+        if np.random.rand() > 0.5:
```
fepegar, line 198: What is this for?

tests/utils.py
```diff
@@ -168,3 +169,34 @@ def assertTensorEqual(self, *args, **kwargs):  # noqa: N802
...
+            ):
+        shape = (*shape, 1) if len(shape) == 2 else shape
+        if no_channel_dim:
+            data = np.random.rand(*shape)
+        else:
+            data = np.random.rand(components, *shape)
+        if binary:
+            data = (data > 0.5).astype(np.uint8)
+            if not data.sum() and force_binary_foreground:
+                data[..., 0] = 1
```
fepegar, line 192: Can we cover this line in the tests?

torchio/data/image.py
```diff
@@ -61,6 +61,11 @@ class Image(dict):
         tensor: If :py:attr:`path` is not given, :attr:`tensor` must be a 4D
             :py:class:`torch.Tensor` or NumPy array with dimensions
             :math:`(C, W, H, D)`.
+        h5DS: If both path and tensor aren't supplied, then an HDF5 Dataset has
```
fepegar, line 64: Can you please rename this variable to something more informative? Maybe `hdf5_dataset`.

torchio/data/image.py
```diff
@@ -61,6 +61,11 @@ class Image(dict):
...
+            to be supplied. 
+        lazypatch: Only applicable with h5DS. If ``True``, the h5DS won't be
```
fepegar, line 66: Can you rename this to snake case? I.e., `lazy_patches`.

torchio/data/image.py
```diff
@@ -97,6 +102,8 @@ def __init__(
             path: Union[TypePath, Sequence[TypePath], None] = None,
             type: str = None,
             tensor: Optional[TypeData] = None,
+            h5DS: Optional[TypeData] = None,
```
fepegar, line 105: `TypeData` means a NumPy array or a PyTorch tensor. Is this annotation correct?

tests/utils.py
```diff
@@ -168,3 +169,34 @@ def assertTensorEqual(self, *args, **kwargs):  # noqa: N802
 
     def assertTensorAlmostEqual(self, *args, **kwargs):  # noqa: N802
         assert_array_almost_equal(*args, **kwargs)
+
+    def get_h5DS_path(
```
fepegar, line 173: What does this do? Can you create a docstring?

setup.py
```diff
@@ -21,6 +21,7 @@
     'torch>=1.1',
     'torchvision',
     'tqdm',
+    'h5py'
```
fepegar, line 24: Maybe this can be added as an optional requirement, as it's more like an extension to the library.

Then we can add it to `requirements-dev.txt` for the tests and raise `ImportError`s if they're needed for whatever the user is trying.

fepegar (changes requested):
Thanks @soumickmj. I have added some comments. I think it might be a good idea to make this a subclass of `Image`, instead of modifying this class so much. What do you think?

.gitignore
```diff
@@ -109,3 +109,4 @@ ENV/
 test_scripts/
 
 *DS_Store
+tester.py
```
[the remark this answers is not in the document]
soumickmj: Sure

tests/utils.py
soumickmj: affine has no use at the moment and can be removed. Rest are same as file creation, to create an HDF4 file inside the temp for tests.

tests/utils.py
soumickmj: Already done. There is a test for it "test_label_map_type_h5ds"

torchio/data/image.py
soumickmj: Sure

torchio/data/image.py
soumickmj: Sure

torchio/data/image.py
soumickmj: Oops sorry. Its not. I will modify accordingly.

tests/utils.py
soumickmj: This creates an HDF5 file inside the temp and returns its path. I will add docstring for the same.

setup.py
soumickmj: Okay.

soumickmj:
@fepegar Thanks for your feedback. I will do the changes accordingly. 
Regarding making this a subclass, originally this was a subclass of Image. But I noticed it was easier to keep it inside the Image to have the same flow as with HDF5, most of the things remain same and can be utilized on it. Also, from tests perspective, I didn't encounter any error. That's why decided to go for it as a part of the main Image class. But if you would like to have it like a Subclass, I will then revert back to the original code of mine. Please let me know what do you feel and I will do it accordingly.

tests/utils.py
soumickmj: I added affine to the test cases as well. Now its completely in use.

setup.py
soumickmj: If its part of the Image class, then it has to be inside setup.py as this will be used for defining the input type for the hdf5_dataset variable (having import inside image class). If we keep it as a subclass of Image, then I can only add the import inside that class.

fepegar: Is it possible to do the `import` within that class?

fepegar:
Yeah I do think it's best to create a new subclass of `Image` instead!

soumickmj:
Okay @fepegar . I will create a subclass using the new TorchIO and create a new pull request.

fepegar:
Thanks!

soumickmj:
@fepegar I have noticed that you have done a lot of changes in the cropping functionality. Earlier the crop function was inside Image class but now its separate. For this H5 extension of the Image class, crop has a bit different code. How would you suggest to proceed? 

> Thanks!

fepegar:
If you liked the old version, you can recover the method and put it in the new subclass, I suppose.

soumickmj:
> If you liked the old version, you can recover the method and put it in the new subclass, I suppose.

Yes, I have the code. That's fine.
But the problem is, from the Patch Pipeline, this method won't be called anymore. Rather the new crop will be called.
Actually I liked the new version, not an issue. Just bit confused about the incorporation.

fepegar:
Oh, ok. Great! FYI: in the beginning, cropping was performed using SimpleITK, but that turned out to be slow so I switched to using that method. I have now refactored the bounds transforms to use pure PyTorch, so they're fast again and we don't need the methods anymore.

soumickmj:
Ah awesome. 
I can add option to the crop.py with a sample check. If the image inside sample is a child of this new subclass, then to use the logic. That's okay?

fepegar:
I'm not sure I understand. Wouldn't that be needed for all transformations?

I assume the reason why you'd like to have this class is to quickly get subvolumes of images saved in the disk. You could also do that with NiBabel: `nib.load(path).dataobj[indices]` but loading only subvolumes is typically not beneficial because we want to apply augmentation to the full-FOV volume. Does that make sense?

soumickmj:
Yes, I understand. 
The difference between the nibabel operation and H5 would be that with H5, everything remains as simple array manipulations. With nibabel indices, it changes a lot.

The reason why I wrote this h5 operation was mainly for patching.
When we are working with huge volumes, and patches are to be created using the patch pipeline, it loads the whole volume in t he RAM. So, we were facing problems frequently with RAM out of memory. So, with H5, when we are loading a patch, it just loads that patch in the RAM.

fepegar:
> The difference between the nibabel operation and H5 would be that with H5, everything remains as simple array manipulations. With nibabel indices, it changes a lot.

Not sure what you mean. These commands to extract a patch as a NumPy array from a huge volume seem very simple (and fast):

```python
In [6]: x = np.zeros((1000, 1000, 1000), np.uint8)                                                                                                                                                                                                                    

In [7]: nib.Nifti1Image(x, np.eye(4)).to_filename('test.nii')                                                                                                                                                                                                         

In [8]: nii = nib.load('test.nii')                                                                                                                                                                                                                                    

In [9]: %timeit nii = nib.load('test.nii')                                                                                                                                                                                                                            
518 µs ± 3.93 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [10]: %timeit patch = nii.dataobj[50:100, 200:300, 1:129]                                                                                                                                                                                                          
23.6 ms ± 40.9 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

soumickmj:
Yes, that is correct.
But that's not what I'm talking about exactly.

I'm talking about patching using https://torchio.readthedocs.io/data/patch_training.html#patchsampler

If we are creating the patches manually in our code, what you said is totally fine.

But when we are using the patch based training option of torchio, we are supplying either image path or tensor. In both cases the complete one. The patches are generated using the torchio pipeline. Right?
And to be precise, using the Crop function inside crop.py
Now in this senario, as far as I understood torchio pipeline, it will be loading the whole volume from the path.
So, if we use H5 dataset, instead of loading the whole thing, we can let the loading done with patches.

Was I able to explain?

So basically, with this subclass, user can supply a HDF5 dataset instead of supplying a tensor or numpy array or a path.
If lazyloading is set to False, then this subclass will load the whole dataset (a complete volume) and then continue the pipeline as it is.
But, if lazyloading is set to True in this subclass, then it will just hold the reference to the dataset but won't load the dataset completely. When the Cropping is performed to load the patch, then it will just be loading the part which we have asked for.

fepegar:
Thanks for the explanation! My issue is that, typically, what happens in the queue is 1) full volume is loaded, 2) augmentation is applied, 3) patches are extracted.

Is this for huge histology images or for medical images?

soumickmj:
The images are acquired at 7T, mostly MRAs. In terms of resolution, they are kinda as high as our colleagues could go. 
But when you now mentioned about the augmentation, it made me realize that I haven't considered for augmentation yet. I considered step 1 and 3 but not 2, as we aren't using augmentations. Somehow we will need to think about the augmentation too.

fepegar:
> The images are acquired at 7T, mostly MRAs. In terms of resolution, they are kinda as high as our colleagues could go.

They must be beautiful :heart: 

> But when you now mentioned about the augmentation, it made me realize that I haven't considered for augmentation yet. I considered step 1 and 3 but not 2, as we aren't using augmentations. Somehow we will need to think about the augmentation too.

There you go, that's what I mean. Let me know if you have any ideas about this. It might be worth it modifying the Queue to apply transforms to patches, or something like that. But it would lead to important changes in the library, so we should discuss that in a new issue, when the time comes.

soumickmj:
> > The images are acquired at 7T, mostly MRAs. In terms of resolution, they are kinda as high as our colleagues could go.
> 
> They must be beautiful ❤️
Thanks for the compliment.
They are. But I can't take any credit for them as I'm mostly working for deep learning part, not doing my own acquisitions so much. For my own acquisitions, they are at 3T and usually for undersampled MRI recon, so not so beautiful.  

> 
> > But when you now mentioned about the augmentation, it made me realize that I haven't considered for augmentation yet. I considered step 1 and 3 but not 2, as we aren't using augmentations. Somehow we will need to think about the augmentation too.
> 
> There you go, that's what I mean. Let me know if you have any ideas about this. It might be worth it modifying the Queue to apply transforms to patches, or something like that. But it would lead to important changes in the library, so we should discuss that in a new issue, when the time comes.

Thanks for pointing it out. I will think about the problem and once I have some idea, I will discuss with you before implementing. 
There is series of wrapper I wrote, for super resolution problems. For this, the input and groundtruth images are of different size. For that wrapper too I haven't thought of the augmentation. But that is not using special h5 or anything, its using the Image class (making subclasses). I will test that with augmentations and then get back to you.

Btw, can I have your email address? Maybe you can drop me testmail at soumick.chatterjee@ovgu.de  
We are working together now on torchio. May be we can also have a meeting someday to discuss other possible collaborations.

fepegar:
Sure! Here it is:
https://github.com/fepegar/torchio/blob/50cd65df4c70f38c30f488c45eb10b6eb27d1645/torchio/__init__.py#L4

soumickmj:
Ah I was so stupid. I missed it.

fepegar:
Not at all, it's not obvious to find it within the repo.

romainVala:
Hi there, 
Having transformation define at the patch level, is technically possible, today, but will it make sense ?...

If you take spatial deformation (affine elastic), then you will induce border effect, (which are not there if you take the all image).
So the only solution I can think of, is to work from a larger patch, performed the transformation and crop. If you take the first patch size large enough, so should obtain very similar effects 

For motion it is more difficult, as you really have non local effect, so I do not expect to have similar effect compare to computing from the all image

fepegar:
They would make sense for things like flipping, blurring or noise. But I'm not sure it's worth the effort.

Issue #344, opened 2020-11-03
change default pading for RandomAffine to minimum


romainVala:
**🚀 Feature**
I notice that by default you choose the minimum as default_pad_value for elastic deformation
whereas the default method is otsu for randomAffine

**Motivation**
it seems to me more natural to take the min, as noise is expected in the image border (at least for brain, in 5 of the 6  border slices )
no ?
and it makes more sense to have the same default for both transform

**Pitch**
by the way we should may be add a default_pad_value as argument in the randomElastic too (but minimum is fine, so may be latter  ...)

**Alternatives**

**Additional context**

fepegar:
> it seems to me more natural to take the min, as noise is expected in the image border (at least for brain, in 5 of the 6 border slices )

Sometimes the minimum is very different from the border, so it creates a gradient. I feel like using the Otsu threshold of the values at the border is the most robust option. But I can't think of an obvious counterexample at the moment...

> and it makes more sense to have the same default for both transform

Yes! I added the padding stuff to RandomAffine as an experiment, but it should also be added to Resample and RandomElasticInformation. Actually, it would be nice to merge these 3 one day.

romainVala:
adding the mean value (of the otsu mask) will also add gradient in the image borders
may be padding stuff is a proper solution (I did not know it was possible, does it allow to choose 'reflect' ? )

When talking of the same default I was thinking of changing `RandomAffine` default `default_pad_value`, which is now 'otsu' whereas it is 'minimum' in `RadomElasticInformation` (without the possibility to change it)

Not sure what you have in mind for merging those 3 transforms, but a first step would be to homogenize them, by adding the `default_pad_value` and the padding stuff if possible (with the same default)

fepegar:
> adding the mean value (of the otsu mask) will also add gradient in the image borders

I guess.

> may be padding stuff is a proper solution (I did not know it was possible, does it allow to choose 'reflect' ? )

Hmm you can always use `Pad` with any mode you want, then apply the affine and then `Crop`.

> When talking of the same default I was thinking of changing RandomAffine default default_pad_value, which is now 'otsu' whereas it is 'minimum' in RadomElasticInformation (without the possibility to change it)
Not sure what you have in mind for merging those 3 transforms, but a first step would be to homogenize them, by adding the default_pad_value and the padding stuff if possible (with the same default)

I agree, this needs to be changed. The idea of merging is avoiding loss of information due to multiple interpolations.

romainVala:
[...]

PR #353, opened 2020-11-16
Add some deterministic transforms

### LINKED ISSUE #142: Pass random parameters to apply_transform()
Hi
Your do save the used coarse_grid in history of the transform, but you also should add num_control_points  (to have a complete description, and the possibility to reproduce it)

Just as a side note (not sure you are interested to it) I implement, apply_given_transform where you set the transform explicitly (to reproduce, the one applied, and saved in history )

Many thanks

### LINKED ISSUE #191: Seeding in Transform, make the same transform all the time

### LINKED ISSUE #208: Reproduce a given transform

### LINKED ISSUE #299: Reverse transform
**🚀 Feature**
Reverse the history of transforms applied to a subject.

**Motivation**

It would be great able to reverse the transforms of a subject especially for spatial transforms.
The idea is to use test-time augmentation (TTA) in order to have a better measure of uncertainty and have more robust predictions.
For reference : https://openreview.net/pdf?id=Byxv9aioz 

**Pitch**
Add a `reverse_transform`  method to the subject class or any other equivalent.
This methode takes a subject or an image as argument, just like transforms.

```python
pipeline = Compose([
    RandomAffine(),
    RandomElasticDeformation()
])
def get_TTA_segmentation:
    augmented_subject = pipeline(subject)
    segmentation = classify_subject(augmented_subject) # segmentation given by the neural network in the augmented subject space
    augmented_subject.reverse_transform(segmentation) # segmentation in the original subject space
```

Obviously some transform such as crop introduce a loss of information and are therefore not 'invertible'.
These case should be carefully handled : should `reverse_transform` of `Crop` raise an error or apply an equivalent `Pad` transform ?

**Alternatives**

No alternative considered.

**Additional context**

### LINKED ISSUE #336: Reorder data back to original orientation
Suppose we have code like this to do a prediction on a new MRI. During training the ToCanonical() transform is used to have all training data in the same orientation.

However, my question, how to get the result (output_image) in the original orientation (the same orientation as the input MRI) instead of the canonical orientation?

```python
    transforms = [torchio.ToCanonical(), ....]

    subject = torchio.Subject(
        ...
    )

    sampler = torchio.inference.GridSampler(
        transform(subject),
        params['patch_size'],
        0, #int(params['patch_size']) / 2,
    )

    patch_loader = torch.utils.data.DataLoader(sampler, batch_size=params['batch_size'])
    aggregator = torchio.inference.GridAggregator(sampler, overlap_mode='average')

    with torch.no_grad():
        for batch_nr, patches_batch in enumerate(patch_loader):
            print(f'Batch {batch_nr}')

            inputs = patches_batch['mr'][torchio.DATA].to(device)
            targets = patches_batch['ct'][torchio.DATA].to(device)

            locations = patches_batch[torchio.LOCATION]
            predictions = model(inputs)

            aggregator.add_batch(predictions, locations)
    output_tensor = aggregator.get_output_tensor()
    output_image = torchio.ScalarImage(tensor=output_tensor.numpy(), affine=patches_batch['mr']['affine'])

    transform = ?
    output_image = transform(output_image)
```

### LINKED ISSUE #355: AffineTransform landmarks segmentation out of frame
I would like to use RandomeAffine transform on the input data of my FCN.

When running affine transform, you always want to make sure that you are applying the same transform to all input images (label, intensity).

I am having the following situation:

Consider one FCN input being a landmark segmentation (tiny blob). After applying the random affine transform, it would end up being somewhere outside the frame. Therefore, all pixel values of the output volume would be 0 and further data augmentation techniques might fail.

If this kind of situation happens, the affine transform (already applied to other input images) should be reverted and a new random transform should be generated applied. This process should be repeated until successfully applied for all. 

Is there a way of doing this with torchio, or do I have to implement this functionality myself (if have done that myself in our local project)?

fepegar:
# Description
This is a refactoring of the reproducibility system of the library, plus adding new features to invert transforms easily.

# Linked issues
Resolves #191.
Resolves #142.
Resolves #208.
Resolves #299.
Resolves #336.
Resolves #355.

# Further issues to take into account
- [x] ~Saving transforms history~
- [x] Reproducing `RandomNoise` (add `seed` to parameters?)
- [x] ~Add a history to the `Image` class, and how to handle that without code duplication?~
- [x] Think about how to handle multichannel images
- [x] Avoid duplication of `inverse()` method
- [x] Consider adding transform to history inside deterministic transforms
- [x] Handle `Compose` and `OneOf` separately

# Progress

## New transforms

- [x] Affine
- [x] BiasField
- [x] Blur
- [x] ~Downsample~
- [x] ElasticDeformation
- [x] Flip
- [x] Gamma
- [x] Ghosting
- [x] LabelsToImage
- [x] Motion
- [x] Noise
- [x] Spike
- [x] Swap

## To update

- [x] Lambda
- [x] Pad
- [x] Crop
- [x] Resample
- [x] ToCanonical
- [x] ZNormalization
- [x] HistogramStandardization
- [x] RescaleIntensity
- [ ] ~CropOrPad~

## Write inverse() method

- [x] Pad
- [x] Crop
- [ ] Resample?
- [ ] ToCanonical?
- [ ] ZNormalization?
- [ ] HistogramStandardization?
- [ ] RescaleIntensity?

# New usage

```python
import torchio as tio

transforms = (
    tio.RescaleIntensity(),
    tio.RandomAffine(),
    tio.RandomBlur(),
    tio.RandomNoise(),
    tio.Flip(),
)

transform = tio.Compose(transforms)

colin = tio.datasets.Colin27()
transformed = transform(colin)
same_transform = transformed.get_composed_history()  # reproducibility
inverted = transformed.apply_inverse_transform()  # invertibility
```

# Checklist

- [x] Breaking change (would cause existing functionality to change)


fepegar:
@romainVala @GReguig

romainVala:
Hi Fernando, 
it seems you plan a big change here, 

I like the deterministic transform idea, it makes it more clear, and explicit, 
but I see you start to remove (or change) a lot of stuff about history, but it is not clear to me what strategy you want to keep. so I need to wait it is becoming more functional, to see if it meet all needs.

Just to be sure, you also have it in mind (Sorry if I repeat mysefl here ..), About reproducibility it is important to be able to rely on a saved text file (or other format) 
My use case, it to run a lot of random transform and save the history (currently to csv) in addition to some metric. After running it massively parallel on a cluster, I want to be able to plot let say, the dice score of my model as a function of any transform parameter (amplitude of the bias, mean displacement, scaling changes ect)
Then I also want to visually check some specific case (for instance a weird outlier) in this case I need to reconstruct the exact same list of transform that have been applied (including non random transform as padding rescal intensity ...)


I thing it should be made explicit in the example scritp for intance something like 

instead of 
```python
transformed = transform(colin)
h = transformed.history[0]
same_transform = transformed.get_composed_history()
```
I prefer the first version you proposed there https://github.com/fepegar/torchio/issues/299#issuecomment-728364977
```python
transform_name, arguments = transformed.history[0]
same_transform = getattr(tio, transform_name)(**arguments)  # TODO: nice wrapper for this
```

arguments, beeing I guess a dict, that can be save and read from disque if need

fepegar:
Hi, @romainVala.

First I want to say I really appreciate your feedback on this. It's very helpful to know about how you use these features. I also want to apologise for undoing some of the work by @GReguig. His PR was very helpful and, although I'm removing some of the code, I'm building on some of his ideas and other code.

> it seems you plan a big change here,

It is indeed going to be quite an important change. I'll bump the version to `v0.18.0` after this and I'll need to write docs and probably update the notebooks. But I think the new functionalities are worth it and that this will make things easier for users.

> I like the deterministic transform idea, it makes it more clear, and explicit,

Thanks, I think so too. I think this separates nicely the random sampling and the functional operations.

> but I see you start to remove (or change) a lot of stuff about history, but it is not clear to me what strategy you want to keep. so I need to wait it is becoming more functional, to see if it meet all needs.

I'll try to post some examples and explanations here.

> Just to be sure, you also have it in mind (Sorry if I repeat mysefl here ..), About reproducibility it is important to be able to rely on a saved text file (or other format)
My use case, it to run a lot of random transform and save the history (currently to csv) in addition to some metric. After running it massively parallel on a cluster, I want to be able to plot let say, the dice score of my model as a function of any transform parameter (amplitude of the bias, mean displacement, scaling changes ect)
Then I also want to visually check some specific case (for instance a weird outlier) in this case I need to reconstruct the exact same list of transform that have been applied (including non random transform as padding rescal intensity ...)

As I said, this type of feedback is very useful. I'll take this into account and bring back some of the `json` functionalities. How do you get the transforms history when you're using the queue?

> I thing it should be made explicit in the example script

I have turned `history` into a higher-level `property` and added `get_composed_history()` to get an even-higher-level way to get a single transform representing the history. But you can still access the old way using the `applied_transforms` attribute:

```python
In [1]: import torchio as tio
   ...: colin = tio.datasets.Colin27()
   ...: tr = tio.RandomAffine()
   ...: c2 = tr(colin)

In [3]: c2.applied_transforms
Out[3]: 
[('Affine',
  {'scales': (0.951589047908783, 0.9638317227363586, 1.0827720165252686),
   'degrees': (-9.776491165161133, -8.459989547729492, 9.02165412902832),
   'translation': (0.0, 0.0, 0.0),
   'center': 'image',
   'default_pad_value': 'otsu',
   'image_interpolation': <Interpolation.LINEAR: 'sitkLinear'>})]

In [4]: c2.history
Out[4]: [<torchio.transforms.augmentation.spatial.random_affine.Affine at 0x7fcb37d149d0>]

In [5]: c2.get_composed_history()
Out[5]: <torchio.transforms.augmentation.composition.Compose at 0x7fcb37873730>
```

I would like to modify the transforms that need interpolation so that they store a string, not an `Interpolation`.

Looking forward to your thoughts!

fepegar:
Now that I think of it: is it really worth the effort to generate a human-readable/jsonable, e.g. JSON or YAML version of the transforms parameters? Can't the history just be serialized and saved using Python's `pickle` or `torch.save`?

romainVala:
No worries about the changes, I do not have a "pythonic" expertise as you do, and this is why I appreciate working on this project where you make the effort to produce a very good open library (documentation test ect ...). 

My concern was just about the final functionality (to reproduce a transform in an independent python execution)
the way you present code example make me think it was working only if you have the python transform object available. Since subject.history is now containing the Transform (or the compose of Transform). I understand it make things easier, but I would like to be able to make the same thing from the parameters
Ok I can now get the parameters, with `applied_transforms` attribute, but how to be sure those parameters will be enough to a new identical transform object ?
don't you thing we should impose to all transform a `set` method (that will reproduce the exact same transform, but given the `applied_transforms` attribute ?

so now we have
```python
c2 = tr(colin)
same_transform = c2.get_composed_history()
```

which need to have the transformed subject python object "alive"

I propose to also make sure we can get the same transform with the attribute strategy : something like that :
```python
parameters = c2.applied_transforms
transform_name = ... 
same_transform = getattr(tio, transform_name)(**parameters)
```

may be it is already just fine, because looking at the new code (for instance RandomAffine.apply_transform) it is already the case, since randomAffine is building an Affine with the parameters an input.

If the plan is to do it for all transform (also the preprocessing one), i think I am 100% happy ...

fepegar:
> If the plan is to do it for all transform (also the preprocessing one), i think I am 100% happy ...

Yes, this is the plan. Will take some work, but at least it will make you 100% happy :smile: 

Could you please let me know how you're saving the history? I suppose at the moment it gets lost when the data loader collates the samples to create a batch...

romainVala:
> Now that I think of it: is it really worth the effort to generate a human-readable/jsonable, e.g. JSON or YAML version of the transforms parameters? Can't the history just be serialized and saved using Python's `pickle` or `torch.save`?

json and co may not be the right choice. 
Actually it may not be the torchio job to do the saving. In my usecase, I need to save the transform params, but also the data path, and other metric score I produce on the fly, so I need to implement it outside torchio, and I am not sure I will need a torchio save method ...

fepegar:
> json and co may not be the right choice.
Actually it may not be the torchio job to do the saving. In my usecase, I need to save the transform params, but also the data path, and other metric score I produce on the fly, so I need to implement it outside torchio, and I am not sure I will need a torchio save method ...

Ok this is good. Remember you can do this to save your transform:

```python
In [3]: import torch

In [4]: torch.save(c2.get_composed_history(), '/tmp/c2_history.pth')

In [5]: c2_history = torch.load('/tmp/c2_history.pth')

In [6]: c2_history
Out[6]: 
Compose(
    <torchio.transforms.augmentation.spatial.random_affine.Affine object at 0x7f4b1a7ce1c0>
)

In [7]: same = c2_history(colin)
```

romainVala:
> Could you please let me know how you're saving the history? I suppose at the moment it gets lost when the data loader collates the samples to create a batch...

why should be it lost ?, actually it is currently working fine, 
in my main training loop
```python
# loader is the pytorch dataloader (where data within subject are concatenate in batch
for i, subject in enumerate(loader, 1):

```
we get the same concatenation for history, if I do 
`subject.get('history') `
I get a list a batch size, containing the history of each batch element 

and the same is working identically with queue (nice no ?)
when I do the save in csv, I do see duplicate line with exact same info since the patch are coming from the same transform subject, but it is just fine ...

fepegar:
Hmm. Aren't you using a custom `collate_fn` when you instantiate the data loader? For example `collate_fn=lambda x: x`?

romainVala:
Oups, you are right I forget this point, but yes fabien modified the pytorch collate function, so that it also include history

romainVala:
using collate_fn=lambda x: x works, but you then loos the nice concatenation done on the tensor, so we prefer to modify the original collate function


the modification was to add:
```python
    elif isinstance(elem, container_abcs.Mapping):
        # The only change to the original function is here:
        # if elem has attribute 'history', then a key 'history' is added to the batch
        # which value is the list of the history of the elements of the batch
        dictionary = {key: history_collate([d[key] for d in batch]) for key in elem}
        if hasattr(elem, 'history'):
            dictionary.update({
                'history': [d.history for d in batch]
            })
        return dictionary
```

if someone is interested, it is there https://github.com/romainVala/torchQC/blob/7e509098e433aad2e5d32bcb40cb502d54f9e5df/segmentation/collate_functions.py#L13

romainVala:
great, you are doing it fast !

I wonder, if on should handle, all call to random, the same way as you do it for RandomNoise (ie: adding a seed as transform parameters)
I thing specifically of `BiasField` and `ElasticDeformation`,
There is 2 advantage to it,
_first is avoid storing big matrix in history (even if biasfield coefficients, is not big, Elastic control_point is: 7*7*7*3=1029, not that small ...)
_The second, is that they are less informative than the parameter used for there generation :
`num_control_points` and `max_displacement` are more informative (in term of expected deformation, than the 1029 control_point grids ...
same for the bias, I would prefer to see as input parameter,  `order` and `coefficients_range`  
...

fepegar:
> great, you are doing it fast !

I might have to slow down soon...

> _first is avoid storing big matrix in history (even if biasfield coefficients, is not big, Elastic control_point is: 777*3=1029, not that small ...)

The number of bias field coefficients with default parameters is 20.
For the elastic deformation, I would argue that 4 KB is a very small size nowadays.

> _The second, is that they are less informative than the parameter used for there generation :
num_control_points and max_displacement are more informative (in term of expected deformation, than the 1029 control_point grids ...
same for the bias, I would prefer to see as input parameter, order and coefficients_range
...

I think the point of having these new transforms is that you know exactly what's going on, given certain parameters. Also, the typical usage won't be to call them directly, I'd say. And if one does, they should have the responsibility of correctly formatting the input arguments.

`num_control_points` is just the size of the control points grid, and `max_displacement` is used to generate the random vectors at each control point. I'm not sure these are more informative.

fepegar:
> if someone is interested, it is there https://github.com/romainVala/torchQC/blob/7e509098e433aad2e5d32bcb40cb502d54f9e5df/segmentation/collate_functions.py#L13

Thanks! I wonder if there's a way to get something like this without copying the original code. I tried putting `history` as a key of `Subject` in the `master` branch, but I got an empty `list` in `batch['history']` :(

fepegar:
> I wonder if there's a way to get something like this without copying the original code

This seems to work:

```python
from torch.utils.data._utils.collate import default_collate

def history_collate(batch):
    elem = batch[0]
    if isinstance(elem, tio.Subject):
        dictionary = {key: default_collate([d[key] for d in batch]) for key in elem}
        if hasattr(elem, 'applied_transforms'):
            dictionary.update({
                'applied_transforms': [d.applied_transforms for d in batch]
            })
        return dictionary
```

torchio/data/subject.py
```diff
@@ -119,6 +119,12 @@ def get_composed_history(self):
         from ..transforms.augmentation.composition import Compose
         return Compose(self.history)
 
+    def get_undo_transform(self):
+        return self.get_composed_history().inverse()
+
+    def undo_transform(self):
+        return self.get_undo_transform()(self)
+
```
romainVala, line 127: that is a funny feature, we get from this PR ! well done !
I just wonder if we should call it apply_inverse_transform(). 
"undo" make you thing that will come back at the exact previous values ... but this will not be the case (depending on the transform)

torchio/transforms/augmentation/composition.py
```diff
@@ -41,12 +41,15 @@ def is_invertible(self):
...
-        return Compose(reversed(transforms))
+        transforms = []
+        for transform in self.transforms:
+            if transform.is_invertible():
+                transforms.append(transform.inverse())
+            else:
+                message = f'Skipping {transform.name} as it is not invertible'
+                raise UserWarning(message)
+        transforms.reverse()
+        return Compose(transforms)
```
romainVala, line 57: so you can still do the inverse although the `is_invertible()` methode will return false ?
then when `is_invertible` method will be used ?

torchio/transforms/augmentation/spatial/random_affine.py
```diff
@@ -104,7 +104,7 @@ def __init__(
             translation: TypeOneToSixFloat = 0,
             isotropic: bool = False,
             center: str = 'image',
-            default_pad_value: Union[str, float] = 'otsu',
+            default_pad_value: Union[str, float] = 'minimum',
```
romainVala, line 106: ah qu'est ce que ca fait la ...
-;)
merci

torchio/transforms/transform.py
```diff
@@ -64,35 +64,39 @@ def __call__(
...
 
+    def __repr__(self):
+        if hasattr(self, 'args_names'):
+            names = self.args_names
+            args_strings = (f'{arg}={getattr(self, arg)}' for arg in names)
+            args_string = ', '.join(args_strings)
+            return f'{self.name}({args_string})'
+        else:
+            return super().__repr__()
+
```
romainVala, line 101: just for my understanding, when is this __repr__ called ?

torchio/transforms/augmentation/intensity/random_ghosting.py
```diff
@@ -37,7 +37,6 @@ class RandomGhosting(RandomTransform, IntensityTransform):
             :math:`k`-space center should be restored after removing the planes
             that generate the artifact.
         p: Probability that this transform will be applied.
-        seed: See :py:class:`~torchio.transforms.augmentation.RandomTransform`.
```
romainVala, line 40: I remember getting lost with this seed parameter, deterministic transforms is definitively a better answer (than this previous seed argument)!

romainVala:
Hi fernando, 
it is looking good !

a small comment about the `is_invertibIe` method in  transform.py
It seems that the attribute invert_transform is missing in : random_flip, crop, pad ...  (and remove     def is_invertible(): from radom_flip)

More generally I wonder, which transform can not be inverted

So the transform I am sure they can not be inverted,
`blur`, 
`random_labels_to_image`: we all working on this inversion, by training network to segment, but this is out of scope here ! this transform is a bit special since it is creating a new image (not really transforming the existing one)

transform that could be inverted, (although of course you will loos a lot of information )
_`resample `
if you save the matrix dim, and affine of the space before resampling, you can then do an inverse resampling (pushing back the data in the original space) no ?
_  `downsample` idem
_ `CropOrPad`, idem if you save the original matrix size ...
_ `toCanonical`, idem if you store the original order ...

`_random_spike`, `random_ghosting`, `motion`,  although it may be a little bit tricky to implement, but it is doable (I made it some time ago (for correction purpose) with my motion implementation, and I was surprise to see that for translation you do not loose any information (perfect recover) I will be curious to see if it is still the case for the image domain approach implemented in torchio

preprocessing.intensity (not sure, the strikethrough the word means it will not be implemented ?

`ZNormalization`, `RescaleIntensity` : the mathematical computation can be easily inverted no ? (but you need to store the percentile, mean and std found in the original data (which can be useful by the way)
HistogramStandardization not sure of this one, but I would bet it is the same

romainVala:
**What for ?**
you push the inversion further than the original need mentioned in  #299
In the test time (inversion) augmentation it was only the label that need to be inverted, here you implemented it for (soon ) all transform .
So I just wonder if you have application in  mind ?

I have to think of it, (and I hope we will find some new one). For now, the use I see is to quite anecdotal : to estimate the information loss, (or the difference) induce by applying a transform and its inverse

torchio/transforms/intensity_transform.py
```diff
@@ -10,3 +10,14 @@ def get_images(sample):
...
+    def arguments_are_dict(self):
+        args = [getattr(self, name) for name in self.args_names]
+        are_dict = [isinstance(arg, dict) for arg in args]
+        if all(are_dict):
+            return True
+        elif not any(are_dict):
+            return False
+        else:
+            message = 'Either all or none of the arguments must be dicts'
+            raise ValueError(message)
```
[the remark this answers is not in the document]
fepegar: Because spatial transforms are applied to all images at the same time, whereas you might want to specify different arguments for different images in intensity transforms.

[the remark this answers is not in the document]
fepegar: For example, you might want to add a lot of noise to your CT but not to your MRI, whereas you typically would like the same transform applied to both (or maybe not, for DL applied to registration? Rethinking this would mean another major change and probably means a new issue + PR).

torchio/transforms/augmentation/composition.py
fepegar: I was discussing this with @pedrob37 yesterday. If you apply a composition of e.g. affine and blur and infer a segmentation, you'd like to invert the affine only because 1) what you want is coming back to the original space and 2) it's not possible to undo the blur anyway. I think it's still nice to be able to call this method, getting a warning.

torchio/transforms/augmentation/spatial/random_affine.py
fepegar: Haha maybe this should go into a different PR, I'm mixing too much here. #344 should be addressed somewhere else!

torchio/transforms/transform.py
fepegar: When you print a transform.

```python
In [1]: import torchio as tio
   ...: colin = tio.datasets.Colin27()
   ...: transform = tio.Compose([tio.RescaleIntensity(), tio.RandomFlip(), tio.RandomBlur(), tio.CropOrPad((100, 200, 300))])
   ...: transformed = transform(colin)

In [2]: transformed.get_composed_history()
Out[2]: 
Compose(
    RescaleIntensity(out_min_max=(0, 1), percentiles=(0, 100), masking_method=<function NormalizationTransform.ones at 0x7f4da759bdc0>)
    Flip(axes=(0,))
    Blur(std={'t1': [tensor([0.2149, 1.7382, 1.0895])]})
    Pad(padding=(0, 0, 0, 0, 60, 59), padding_mode=constant)
    Crop(cropping=(41, 40, 9, 8, 0, 0))
)

In [3]: transformed.get_inverse_transform()
/home/fernando/git/torchio/torchio/transforms/augmentation/composition.py:56: UserWarning: Skipping RescaleIntensity as it is not invertible
  warnings.warn(message, UserWarning)
/home/fernando/git/torchio/torchio/transforms/augmentation/composition.py:56: UserWarning: Skipping Blur as it is not invertible
  warnings.warn(message, UserWarning)
Out[3]: 
Compose(
    Pad(padding=(41, 40, 9, 8, 0, 0), padding_mode=constant)
    Crop(cropping=(0, 0, 0, 0, 60, 59))
    Flip(axes=(0,))
)
```

I still need to figure out if there's a nice way to do this with random transforms.

torchio/transforms/augmentation/intensity/random_ghosting.py
fepegar: I think so too! At least it's more explicit (which is better than implicit).

```python
import this
```

fepegar:
Many thanks for reviewing, Romain!

> a small comment about the is_invertibIe method in transform.py
It seems that the attribute invert_transform is missing in : random_flip, crop, pad ... (and remove def is_invertible(): from radom_flip)

The method is inherited from `Transform`:

```python
def is_invertible(self):
    return hasattr(self, 'invert_transform')
```

> So the transform I am sure they can not be inverted,
blur,
random_labels_to_image: we all working on this inversion, by training network to segment, but this is out of scope here ! this transform is a bit special since it is creating a new image (not really transforming the existing one)

Agree.

> transform that could be inverted, (although of course you will loos a lot of information )

> _resample
if you save the matrix dim, and affine of the space before resampling, you can then do an inverse resampling (pushing back the data in the original space) no ?
_ toCanonical, idem if you store the original order ...

Yeah I need to give these a thought. I suppose it would mean adding an additional kwarg to the transforms, such as "original affine" or so. I'm not too concerned about the reorientation, but inverting a resampling would be useful after e.g. a RandomDownsample, which was my original idea when I created the transform (of course, Benjamin Billot confirmed this with his paper!).

> _ downsample idem
_ CropOrPad, idem if you save the original matrix size ...

These two call other transforms that will be added to the history, so no extra work needs to be done.

> _random_spike, random_ghosting, motion, although it may be a little bit tricky to implement, but it is doable (I made it some time ago (for correction purpose) with my motion implementation, and I was surprise to see that for translation you do not loose any information (perfect recover) I will be curious to see if it is still the case for the image domain approach implemented in torchio

I'm not sure it's worth it to work on inverting these. I think it's fine to have some transforms that are invertible and some that are not. Unless we can come up with a good use case to invert them. I think that inverting e.g. RandomMotion is going to be painful and doesn't feel like it would help much.

> preprocessing.intensity (not sure, the strikethrough the word means it will not be implemented ?
ZNormalization, RescaleIntensity : the mathematical computation can be easily inverted no ? (but you need to store the percentile, mean and std found in the original data (which can be useful by the way)
HistogramStandardization not sure of this one, but I would bet it is the same

My feeling about this is a mix of my two comments above: 1) would need to change the design and 2) not sure it's worth the effort.

fepegar:
> What for ?
you push the inversion further than the original need mentioned in #299
In the test time (inversion) augmentation it was only the label that need to be inverted, here you implemented it for (soon ) all transform .
So I just wonder if you have application in mind ?
I have to think of it, (and I hope we will find some new one). For now, the use I see is to quite anecdotal : to estimate the information loss, (or the difference) induce by applying a transform and its inverse

I'm not sure I understand. If only the label needs to be inverted (I assume you mean the result of a segmentation inference), the other images in the subject could be deleted, no?

torchio/transforms/augmentation/composition.py
romainVala: Yes I agree,

romainVala:
Ok I agree, 

the "What for" question, was just a general one, to seek for argument, again what I was expected you will answer : 

>  2) not sure it's worth the effort.

Learning with synthetic data, as propose, by billot, open so much new perspective, that I have the feeling it worth the effort. (as you point it for resampling). we just do not see it today ...

about change in design, yes, but not that much, and since you are changing a lot here, I though it may be the right time

But ok, let's be pragmatic,  we can address it later
I'll come back later with new issue, when I really need it.

About inverting motion, I will hope I'll find time to test it, but I think it may help for learning a motion correction method. (at least my motion implementation make it easy to invert, so I can already test it ...)

fepegar:
Maybe I'm failing to understand a use case, probably because I'm strongly biased towards 1) brain 2) MRI 3) segmentation. But as you say, let's be pragmatic. I'd rather push for an "incomplete" feature than investing a lot of time so that the number of invertible transforms in the library is as large as possible. We can always add the functionality later on! This PR will provide the infrastructure to do that easily.

romainVala:
Congratulation !
You did well for this  "good first issue"  
-;)

fepegar:
Haha thanks Romain! And as usual, thanks a lot for your feedback! I hope this PR is useful for you.

PR #369, opened 2020-12-05
Fixes to internal references in docs

### LINKED ISSUE #247: Some links to classes in docs are not working
For example, in [`RandomLabelsToImage`](https://torchio.readthedocs.io/transforms/augmentation.html#randomlabelstoimage), the link to `RandomTransform` works but not the one to `RandomBlur`.

jwitos:
Fixes #247.

**Description**
As discussed in #247, I fixed links to classes throughout the docs. As you suggested, Sphinx has some issues regarding to namespaces and modules. I kept the `.. currentmodule::`, and as long as you specify the module in the doc file, e.g. `.. currentmodule:: torchio.data`, I think you should be able to refer to classes within this module in other doc files/docstrings in the form of ``:class:`torchio.data.ClassName` ``, at least I was able to.

The only problem major problem with Sphinx was that it shows Base class inheritance inconsistently. For instance, if you inherit `SubjectsDataset`, it will say that the Base class is `torchio.data.dataset.SubjectsDataset` instead of `torchio.data.SubjectsDataset`, breaking the link. To work around this, I have added in some files a `__module__` attribute to imported classes with incorrect paths, e.g. `SubjectsDataset.__module__ = "torch.utils.data"`. This fixes the Base class, but leaves slightly annoying line of code. I can get rid of those `__module__` assignments, but then links to Base classes won't work. (I'm sure that there is another way to do this but I can't figure it out)

**Checklist**

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -54,6 +54,7 @@ intersphinx_mapping = {
     'torch': ('https://pytorch.org/docs/master/', None),  # https://github.com/pytorch/fairseq/blob/adb5b9c71f7ef4fe2f258e0da102d819ab9920ef/docs/conf.py#L131
     'torchvision': ('https://pytorch.org/docs/master/', None),
     'nibabel': ('https://nipy.org/nibabel/', None),
+    'numpy': ('https://numpy.org/doc/stable/', None),
 }
 
 # Add any paths that contain templates here, relative to this directory.
--- a/docs/source/data/image.rst
+++ b/docs/source/data/image.rst
@@ -2,7 +2,7 @@
 Image
 #####
 
-The :class:`Image` class, representing one medical image,
+The :class:`~torchio.data.Image` class, representing one medical image,
 stores a 4D tensor, whose voxels encode, e.g., signal intensity or segmentation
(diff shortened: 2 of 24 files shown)


fepegar (commented):
I added a couple of comments. I don't think modifying a the `__module__` is a good idea, there must be a better solution for this...

> The only problem major problem with Sphinx was that it shows Base class inheritance inconsistently. For instance, if you inherit SubjectsDataset, it will say that the Base class is torchio.data.dataset.SubjectsDataset instead of torchio.data.SubjectsDataset, breaking the link.

I'm not sure why it should be `torchio.data.SubjectsDataset` instead of `torchio.data.dataset.SubjectsDataset`, which is where the class is actually defined.

Can't the `.. currentmodule::` directive be used to fix these issues?

fepegar:
Simply adding `.. currentmodule:: torchio` at the top of most `rst` files seems to work. If you don't mind, I'll try that in a new PR and then we can work from there.

fepegar:
I refactored quite a lot of stuff in the docs, links seem to be working now. Your investigation was very helpful, @jwitos. Thanks a lot for your contribution!

PR #370, opened 2020-12-05
Replacing p by **kwargs in all transforms

### LINKED ISSUE #359: Support for applying intensity transforms to a subset of images in the Subject
**🚀 Feature**
At the moment an intensity transform is applied to all images of type intensity. There are use cases when a certain normalization/noise adding/clipping should only apply to some images in the subject

I was looking in the code and I think this feature could be achieved by changing the IntensityTransform to 

```python
from typing import Optional, List

from .transform import Transform
from .. import Image


class IntensityTransform(Transform):

    def __init__(
        self,
        images_to_apply: Optional[List[str]] = None,
        p: float = 1,
        copy: bool = True,
        keys: Optional[List[str]] = None,
    ):
        super().__init__(p=p, copy=copy, keys=keys)
        self.images_to_apply = images_to_apply

    """Transform that modifies voxel intensities only."""
    #@staticmethod
    def get_images(self, sample):

        images_dict = self.get_images_dict(sample)
        return list(images_dict.values())

        #return sample.get_images(intensity_only=True)

    #@staticmethod
    def get_images_dict(self, sample):
        images = {}
        for image_name, image in sample.items():
            if not isinstance(image, Image):
                continue
            if self.images_to_apply is not None and image_name not in self.images_to_apply:
                continue
            images[image_name] = image
        return images

        #return sample.get_images_dict(intensity_only=True)

```

Note the extra parameter `images_to_apply`. In this way all intensity transforms do only get the images they should transform. Is this the right way to approach this or am I missing something? 

When you agree with this approach I can try to make a pull request of it

dmus:
Replacing p by **kwargs in all transforms

First step to fix #359 .

**Description**
This initial pull request adds **kwargs to all transforms to be able to use include and exclude parameters later on. Eventually this work can first be merged. Then the include and exclude will follow in a new PR.

**Checklist**

- [x ] I have read the [`CONTRIBUTING`](https://github.com/fepegar/torchio/blob/master/CONTRIBUTING.rst) docs and have a developer setup (especially important are `pre-commit`and `pytest`)
- [x ] Non-breaking change (would not break existing functionality)
- [ x] Integration tests passed locally by running `pytest`
- [x ] In-line docstrings updated
- [ x] Documentation updated, tested running `make html` inside the `docs/` folder
- [ x] This pull request is ready to be reviewed


fepegar:
I fixed some conflicts introduced by #371.

Could you please install `pre-commit` as explained in [CONTRIBUTING](https://github.com/fepegar/torchio/blob/master/CONTRIBUTING.rst)? I am getting some flake8 errors that shouldn't have allowed you to commit if you had it installed correctly.

dmus:
@fepegar pre-commit is installed succesfully now and I updated the PR. Could this PR first be merged, followed by a new PR for the include/exclude, or shall I continue with that on this branch?

I see that there are two checks about code coverage not successful because two import lines were removed. I assume that is not a problem?

fepegar:
Yes, let's not worry about those coverage tests now. I'll merge in a second.

fepegar:
@all-contributors please add @dmus for doc, code

fepegar:
Thanks for your contribution, @dmus!

PR #373, opened 2020-12-07
Implementation of include/exclude args

### LINKED ISSUE #54: Add exclude/include parameters to transforms kwargs
Use case: loading an MRI and CT from the same patients. We probably want very different transforms for each image, but we still want them to happen in the same parallel process.

dmus:
Fixes #54 .

**Description**
This proposes the use of include and exclude args for transforms. In this way a transform can be applied to a subset of the images in a dictionary, needed when some image need other preprocessing/augmentation than others

**Checklist**

### DIFF
--- a/torchio/transforms/intensity_transform.py
+++ b/torchio/transforms/intensity_transform.py
@@ -1,4 +1,4 @@
-from typing import List, Dict
+from typing import Dict
 
 from .transform import Transform
 from ..data import Image
@@ -8,9 +8,6 @@ class IntensityTransform(Transform):
     """Transform that modifies voxel intensities only."""
     def get_images_dict(self, sample) -> Dict[str, Image]:
         return sample.get_images_dict(intensity_only=True, include=self.include, exclude=self.exclude)
-
-    def get_images(self, sample) -> List[Image]:
-        return sample.get_images(intensity_only=True, include=self.include, exclude=self.exclude)
 
     def arguments_are_dict(self) -> bool:
         """Check if main arguments are dict.
--- a/torchio/transforms/intensity_transform.py
+++ b/torchio/transforms/intensity_transform.py
@@ -1,7 +1,7 @@
 from typing import List, Dict
 
 from .transform import Transform
-from .. import Image
+from ..data import Image
 
 
 class IntensityTransform(Transform):
--- a/torchio/transforms/intensity_transform.py
+++ b/torchio/transforms/intensity_transform.py
@@ -1,11 +1,16 @@
+from typing import List, Dict
+
 from .transform import Transform
+from .. import Image
 
 
 class IntensityTransform(Transform):
     """Transform that modifies voxel intensities only."""
-    @staticmethod
-    def get_images_dict(sample):
-        return sample.get_images_dict(intensity_only=True)
+    def get_images_dict(self, sample) -> Dict[str, Image]:
+        return sample.get_images_dict(intensity_only=True, include=self.include, exclude=self.exclude)
+
+    def get_images(self, sample) -> List[Image]:
+        return sample.get_images(intensity_only=True, include=self.include, exclude=self.exclude)
 
     def arguments_are_dict(self) -> bool:
         """Check if main arguments are dict.
--- a/torchio/transforms/spatial_transform.py
+++ b/torchio/transforms/spatial_transform.py
@@ -1,4 +1,4 @@
-from typing import Dict, List
+from typing import List
 
 from .transform import Transform
 from ..data import Image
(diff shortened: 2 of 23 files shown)


dmus:
@fepegar I have a question about how the imports are organized. I have never used relative imports myself. In spatial_transform.py I have this `from .. import Image` (suggested by PyCharm) but this results in an error when running pytest:

> E   ImportError: cannot import name 'Image'

How can this be solved?

fepegar:
Thanks for working on this, @dmus. Is this ready for review? If not, could you please mark it as a draft PR?

---

> How can this be solved?

I suspect that's because of circular imports, i.e. spatial transform importing image and image importing spatial transform, maybe indirectly. I am not very good at these and suspect that using absolute imports everywhere would help. But that's another issue. I think in this case you can try importing the class from its actual module:

```python
from ..data.image import Image
```

I tried `from .. import Image` and got:

```python-traceback
ImportError: cannot import name 'Image' from partially initialized module 'torchio' (most likely due to a circular import) (/Users/fernando/git/torchio/torchio/__init__.py)
```

which is a nicer error message, maybe because I'm using a newer version of Python.

---

One more thing about markdown. This code

```
- [x] Correct
- [ ] Correct
- [] Incorrect
- [x ] Incorrect
- [ x] Incorrect
```

generates this:

- [x] Correct
- [ ] Correct
- [] Incorrect
- [x ] Incorrect
- [ x] Incorrect

dmus:
@fepegar (or someone else who has good suggestions), could you have a look at line 48 and 49 of random_noise.py:

   ```
 def apply_transform(self, subject: Subject) -> Subject:
        arguments = defaultdict(dict)
        for image_name in self.get_images_dict(subject):
            mean, std, seed = self.get_params(self.mean_range, self.std_range)
            arguments['mean'][image_name] = mean
            arguments['std'][image_name] = std
            arguments['seed'][image_name] = seed
        arguments['include'] = self.include
        arguments['exclude'] = self.exclude
        transform = Noise(**arguments)
        return transformed
```

For the random transforms composition is preferred over inheritance. However, is there some solution possible that we don't have to add these two lines `arguments['include'] = self.include` and `arguments['exclude'] = self.exclude` in all random transforms?

**Comment by fepegar** [comment]
I can't think of a much better solution.

If you want a quick search & replace solution, maybe we can add a method in `RandomTransform` to update `arguments` with `include` and `exclude`:

```python
def add_include_exclude(self, kwargs):
    kwargs['include'] = self.include
    kwargs['exclude'] = self.exclude
    return kwargs
```

and then replace all `(**arguments)` with `(**self.add_include_exclude(arguments))`. What do you think?

fepegar:
[Just in case: `**arguments` is also present in `Subject.history` and that one should not be modified]

dmus:
Okay, that seems like a reasonable solution. I implemented that. Please review this PR. I was not completely sure how to organize the tests, if there are more/other tests needed let me know

torchio/transforms/intensity_transform.py
```diff
@@ -1,11 +1,16 @@
...
 
 class IntensityTransform(Transform):
     """Transform that modifies voxel intensities only."""
-    @staticmethod
-    def get_images_dict(sample):
-        return sample.get_images_dict(intensity_only=True)
+    def get_images_dict(self, sample) -> Dict[str, Image]:
+        return sample.get_images_dict(intensity_only=True, include=self.include, exclude=self.exclude)
+
+    def get_images(self, sample) -> List[Image]:
```
fepegar, line 12: I don't think this method is necessary.

torchio/transforms/spatial_transform.py
```diff
@@ -1,12 +1,14 @@
...
 class SpatialTransform(Transform):
     """Transform that modifies image bounds or voxels positions."""
-    @staticmethod
-    def get_images(sample):
-        return sample.get_images(intensity_only=False)
 
-    @staticmethod
-    def get_images_dict(sample):
-        return sample.get_images_dict(intensity_only=False)
+    def get_images_dict(self, sample) -> Dict[str, Image]:
```
fepegar, line 10: I'm not sure this is necessary either.

fepegar (changes requested):
It looks pretty good. I added some comments and committed some minor changes.

torchio/transforms/intensity_transform.py
dmus: @fepegar is there a reason that in spatial transforms (for example random_affine.py) this is used:

    ```
def apply_transform(self, subject: Subject) -> Subject:
        ...
        for image in self.get_images(subject):
```

and in intensity transforms (for example random_blur.py):
    ```
def apply_transform(self, subject: Subject) -> Subject:
        ...
        for name, image in self.get_images_dict(subject).items():
```
Could we make this consistent?

fepegar: The reason is that in spatial transforms you don't need to know the name of images because the same parameters are applied to all of them, whereas for intensity transforms you retrieve the name to store it in the parameters dict. Does that make sense?

dmus: okay makes sense, will remove methods

fepegar:
Added in `v0.18.2`. Thanks for your contribution, @dmus!

PR #380, opened 2020-12-12
Remove checks and casts to float32

### LINKED ISSUE #375: Other datatype for LabelMap than float32
**🚀 Feature**
I noticed that a LabelMap and an IntensityImage are both saved as float32 tensors, which means that the LabelMap uses a lot more memory than needed. This is because this piece of code in io.py which casts all image to float32:

```
def _read_sitk(path: TypePath) -> Tuple[torch.Tensor, np.ndarray]:
    if Path(path).is_dir():  # assume DICOM
        image = _read_dicom(path)
    else:
        image = sitk.ReadImage(str(path))
    data, affine = sitk_to_nib(image, keepdim=True)
    if data.dtype != np.float32:
        data = data.astype(np.float32)
    tensor = torch.from_numpy(data)
    return tensor, affine
```

Is there a reason for this .astype(np.float32)?

This can be made a lot more memory friendly by removing this cast and storing segmentations in memory as uint8 for example. Also I expect spatial augmentations which requires resampling to be a lot faster when they work with uint8 instead of float32

**Motivation**

- Better use of memory
- Faster augmentations which require resampling

**Pitch**

No cast to float32 for all tensors, allowing different dtypes

Could these two lines be removed? All tests still pass when I comment them out. Maybe only cast bool to np.uint8 because SimpleITK does not support bool?
```
 if data.dtype != np.float32:
        data = data.astype(np.float32)
```

dmus:
Fixes #375 .

**Description**
Remove of the casts and checks for float32 to allow other datatypes (uint8 for label masks for example)

**Checklist**

### DIFF
--- a/tests/data/test_image.py
+++ b/tests/data/test_image.py
@@ -9,9 +9,8 @@ import torch
 import numpy as np
 import nibabel as nib
 
-from torchio import ScalarImage, LabelMap, Subject, INTENSITY, LABEL, STEM
+import torchio as tio
 from ..utils import TorchioTestCase
-from torchio import RandomFlip, RandomAffine
 
 
 class TestImage(TorchioTestCase):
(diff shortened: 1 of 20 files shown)


dmus:
I removed the explicit casts and checks for float32. I found I had to add a cast to float in label.py and in random_labels_to_image.py

I changed `self.assertTensorEqual` to `self.assertTensorAlmostEqual` in test_canonical. It turns out the test data is in float64, another option is to have the test data in float32, such that the tests runs as before

fepegar:
Thanks for working on this. Here's a problem introduced by these changes, when reading an image stored in `uint16`: `TypeError: can't convert np.ndarray of type numpy.uint16. The only supported types are: float64, float32, float16, complex64, complex128, int64, int32, int16, int8, uint8, and bool.`.

dmus:
Do you have some code to reproduce? How to solve that?

fepegar:
> Do you have some code to reproduce?

```python
import torch
import numpy as np
x = np.array(1, np.uint16)
torch.from_numpy(x)
```

```python-traceback
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-1-084a661a5760> in <module>
      2 import numpy as np
      3 x = np.array(1, np.uint16)
----> 4 torch.from_numpy(x)

TypeError: can't convert np.ndarray of type numpy.uint16. The only supported types are: float64, float32, float16, complex64, complex128, int64, int32, int16, int8, uint8, and bool.
```

> How to solve that?

I suppose the images should be cast to a data type that can represent the original values. I suppose a good way to do this is casting uint16 to int32 and uint32 to int64. What do you think?

dmus:
Yes I think that is a good solution

dmus:
I also added a cast from bool to uint8, because bool is not supported by SimpleITK

fepegar:
I have added some necessary casts here and there, and updated the tests. Shall we merge?

dmus:
Good work. Yes, let's merge

fepegar:
I added some more checks that were causing loading of uint16 to fail. Thanks for your contribution, @dmus!

PR #398, opened 2020-12-27
Fix error with invertible custom transforms, also keep include/exclude lists for inverse transforms.

efirdc:
**Description**
This fixes two bugs that I found when making a custom invertible `Transform`.

First is that `Subject.history` would not find any transforms defined outside torchio. To fix this I added a `get_subclasses(target_class)` function to `utils.py`.

Other thing I noticed was that `Transform._get_reproducing_arguments()` did not keep the include/exclude lists. So I added them there.

**Checklist**

- [x] If the PR is ready and there are multiple commits, I have attempted to [squash them and force-push](https://www.w3docs.com/snippets/git/how-to-combine-multiple-commits-into-one-with-3-steps.html#force-pushing-commits-7)

### DIFF
--- a/torchio/data/subject.py
+++ b/torchio/data/subject.py
@@ -6,6 +6,7 @@ import numpy as np
 
 from ..constants import TYPE, INTENSITY
 from .image import Image
+from ..utils import get_subclasses
 
 
 class Subject(dict):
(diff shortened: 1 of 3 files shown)


fepegar:
Nice. The invertibility feature is very young and I'm sure there is still a lot of room for improvement. Many thanks for your contribution, @efirdc!

fepegar:
@all-contributors please add @efirdc for code

romainVala:
Hello
just out of curiosity, @efirdc do you use transform outside torchio (which one, how it is compatible with torchio) ?

efirdc:
Hi @romainVala,

At the time it was the label transformations, but they are included in torchio now. At the moment I'm not using any custom transformations.

Issue #400, opened 2020-12-27
Label transformations


efirdc:
## **🚀 Feature**

Currently working on a `RemapLabels` transformation for `LabelMap`s as well as two derivative transformations: `RemoveLabels` and `SequentialLabels` . Would be happy to submit a PR for this if there is interest.

`RemapLabels` takes a `remapping: Dict[int, int]` which specifies how labels should be remapped. This only changes the underlying integer IDs of a class. For example `remapping={20:1, 2:3, 3:2, 4:0}` would remap 20 to 1, swap 2 and 3, and remove 4 by setting it to background.

`SequentialLabels` will make all labels sequential. For example if you have a label map with labels `2, 9, 10, 15, 16`, this will remap them to `1, 2, 3, 4, 5`. This transform would just apply a `RemapLabels` transform with `remapping={2:1, 9:2, 10:3, 15:4, 16:5}`.

`RemoveLabels` takes a list of labels and maps them all to zero. This will also reuse `RemapLabels`. It could take an option to specify which label is considered background. Although I doubt anyone would use something other than zero. 

Invertibility is easily supported for `RemapLabels` by swapping keys and values for `remapping`. This would also automatically make `SequentialLabels` invertible as well. The main purpose of `RemoveLabels` would be to disable invertibility when remapping classes to background.

There is another more difficult use case I would like to support of combining two labels into a single label, and then separating them again in the inverse pass. This would require an additional `masking_method` parameter.

## **Motivation**

Currently working with a dataset that has quite a few labels for deep gray matter structures:
```python
label_names={'left_ventricle': 1, 'right_ventricle': 2, 'left_caudate': 3, 'right_caudate': 4, 'left_putamen': 5, 'right_putamen': 6,
                      'left_thalamus': 7, 'right_thalamus': 8, 'left_globus_pallidus': 9, 'right_globus_pallidus': 10, 'internal_capsule': 17,
                      'left_red_nucleus': 19, 'right_red_nucleus': 20,'left_substantia_nigra': 21, 'right_substantia_nigra': 22, 
                      'left_dentate_nucleus': 23, 'right_dentate_nucleus': 24}
```
As you can see they are all split into left/right classes and there are a lot of gaps between the label numbers. This is the desired transformation pipeline:

```python
transforms = tio.Compose([
    RemoveLabels([1, 2, 17, 23, 24]),
    RemapLabels({4:3, 6:5, 10:9, 20:19, 22:21})
    SequentialLabels(),
])
```

[...]

PR #402, opened 2020-12-28
New label transformations

efirdc:
Implements #400.

**Description**
This implements new label transformations `RemapLabels`, `SequentialLabels`, `RemoveLabels`. Documentation and basic tests are included as well. Please let me know if anything should be improved or changed.

Some extra methods were added to `TorchioTestCase` for easier testing of these new transformations.
- `get_subject_with_labels(self, labels):` returns a subject with a `LabelMap` which has the desired set of `labels`.
- `get_unique_labels(self, label_map):` returns the set of unique labels in a `LabelMap`, not including 0.

A small fix to `Subject.apply_inverse_transform`.  Just applied `.clear_history()` to the subject after the inverse transformation.

Moved everything related to `TypeBounds`, and `TypeMaskingMethod` up to the `Transform` class so they can be reused in the label transformations. 

**Demonstration**

Original labels:

Apply `tio.RemoveLabels([1, 2, 17, 23, 24])`

Apply `tio.RemapLabels({4:3, 6:5, 10:9, 20:19, 22:21}, masking_method="Right")`

Apply `tio.SequentialLabels()`

Apply `subject = subject.apply_inverse_transform()`

**Checklist**

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -119,6 +119,20 @@ class TorchioTestCase(unittest.TestCase):
             ),
         )
 
+    def get_subject_with_labels(self, labels):
+        return tio.Subject(
+            label=tio.LabelMap(
+                self.get_image_path(
+                    'label_multi', labels=labels
+                )
+            )
+        )
+
+    def get_unique_labels(self, label_map):
+        labels = torch.unique(label_map.data)
+        labels = {i.item() for i in labels if i != 0}
+        return labels
+
     def tearDown(self):
         """Tear down test fixtures, if any."""
         shutil.rmtree(self.dir)
@@ -131,6 +145,7 @@ class TorchioTestCase(unittest.TestCase):
             self,
             stem,
             binary=False,
+            labels=None,
             shape=(10, 20, 30),
             spacing=(1, 1, 1),
             components=1,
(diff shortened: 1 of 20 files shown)


fepegar (commented):
Nice! I really appreciate the effort into the docs and the examples.

Why did you move the bounds stuff into Transform? I haven't seen it in the new transforms.

efirdc:
No worries, glad to contribute.

The bounds code was moved to `Transform` so that `TypeBounds` can be supported as a masking method. You can see where that happens on line 402 of `transform.py`

fepegar:
Ok, thanks. I've added some minor edits, will squash and merge once the tests are done.

tests/utils.py
```diff
@@ -144,6 +159,14 @@ def get_image_path(
             data = (data > 0.5).astype(np.uint8)
             if not data.sum() and force_binary_foreground:
                 data[..., 0] = 1
+        elif labels is not None:
+            data = (data * (len(labels) + 1)).astype(np.uint8)
+            new_data = np.zeros_like(data)
+            for i, label in enumerate(labels):
+                new_data[data == (i + 1)] = label
+                if not (new_data == label).sum():
+                    new_data[..., i] = label
```
fepegar, line 168: Are these lines necessary?

fepegar (changes requested):

tests/utils.py
efirdc: Probably not. I saw on line 160 that if by chance the volume has no ones and `force_binary_background` is enabled then it will add a plane of ones. I thought the idea was to prevent a test from failing by random chance.

Although with the default shape of (10, 20, 30) giving 6000 voxels, the chance that any label will be missing is astronomically low. So this can probably be removed.

fepegar: Oh, I see. I was curious about setting the value of one full slice, but I guess it's not important. Let's leave it, as sometimes the default shape won't be used, so this line might be reached.

fepegar:
Available in `v0.18.13`. Thanks again for your contribution!

tests/utils.py
efirdc: Thats true. Although now that I think about it, it should probably be changed to `new_data.flatten()[i] = label` in a future PR. Otherwise it will go out of bounds if the last dimension of the volume is less than the number of desired labels.

efirdc:
Great! Thanks for your edits and continued support of this project.

PR #404, opened 2021-01-01
Allow empty list as input to Compose.

efirdc:
**Description**

This PR is just a small suggestion to allow empty lists as an input to a transform `Compose`. At the moment an error is raised in this case, but I think it would be nicer to return a null transform. This would be important if you are building a list of transforms dynamically and there is a chance of having an empty list. If the empty list case is not explicitly handled then this will introduce a hidden bug.

I ran into this situation when I was trying to construct a transform inverse, but only including the label transformations:

```python
label_transformations = []
for transform in subject.history:
    if type(transform) in (tio.RemapLabels, tio.RemoveLabels, tio.SequentialLabels):
        label_transformations.append(transform)
inverse_label_transform = tio.Compose(label_transformations ).inverse(warn=False)
```
This code worked until a case where the subject did not have any label transformations applied.

**Checklist**

### DIFF
--- a/torchio/transforms/augmentation/composition.py
+++ b/torchio/transforms/augmentation/composition.py
@@ -23,8 +23,6 @@ class Compose(Transform):
     """
     def __init__(self, transforms: Sequence[Transform], **kwargs):
         super().__init__(**kwargs)
-        if not transforms:
-            raise ValueError('The list of transforms is empty')
         for transform in transforms:
             if not callable(transform):
                 message = (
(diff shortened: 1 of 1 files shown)


torchio/transforms/augmentation/composition.py
```diff
@@ -65,12 +63,9 @@ def inverse(self, warn: bool = True) -> Transform:
...
                 warnings.warn(message, RuntimeWarning)
         transforms.reverse()
-        if transforms:
-            result = Compose(transforms)
-        else:  # return noop if no invertible transforms are found
-            def result(x): return x  # noqa: E704
-            if warn:
-                warnings.warn('No invertible transforms found', RuntimeWarning)
+        result = Compose(transforms)
+        if not transforms:
```
fepegar, line 67: What about `if not transforms and warn`?

fepegar (changes requested):
LGTM. Just a single comment in the code.

torchio/transforms/augmentation/composition.py
efirdc: Whoops! Good catch.

Just pushed a fix.

fepegar:
Thanks, @efirdc. Added in `v0.18.14`.

PR #405, opened 2021-01-01
Add Python 3.9 to CI

fepegar:


fepegar:
The config from the last commit seems to be the correct one. However, building SimpleITK seems to hang on macOS:

To reproduce:

```
10032  1/1/2021 17:32  conda create -n test python ipython -y && act test
10033  1/1/2021 17:33  pip install -r requirements-dev.txt
10034  1/1/2021 17:41  pip install .
10035  1/1/2021 17:41  python -m pip install scikit-build
10036  1/1/2021 17:41  pip install .
```

Issue #345, opened 2020-11-04
strange bug, due to missing copy ... ?


romainVala:
**🐛Bug**

I find python bugs related to missing copy, very difficult to see, I guess this one is related to this ..

when I modify the content of a transformed subject it also change the original subject
This happen only when using LabelMap AND RandomBlur
or to be more precise: it does not happen if on use either scalarImage OR RandomAffine

**To reproduce**

```python
import torchio as tio

colin = tio.datasets.Colin27()
colin.pop('brain')
colin.pop('head')

label_volume = colin.t1.data > 1000000  # pv.argmax(dim=0, keepdim=True)

label = tio.LabelMap(tensor=label_volume, affine=colin.t1.affine)
# label = tio.ScalarImage(tensor=label_volume, affine=colin.t1.affine)

new_subject = tio.Subject(label=label)
# new_subject.plot()

transform = tio.RandomBlur()
# transform = tio.RandomAffine()
transformed = transform(new_subject)

transformed_data = transformed.label.data[0]
transformed_data[:] = 0
```

```python-traceback
# Your errors here
```
it plots a black image, although I only modify the transformed data, 

Now un-comment either RandomAffine, or ScalarImage, and there is no more bug 

**Expected behavior**

**Actual behavior**

**TorchIO version**

romainVala:
note that if I change

`dd=tsub.t1.data[0]`
by
`dd=copy.deepcopy(tsub.t1.data[0])`

then it works, but it should not be necessary ...

fepegar:
Yeah this is related to what is and isn't copied during a transform. It is indeed tricky.

The problem is that copying can be long if the tensor is large, and I'm not sure it's a good idea to copy e.g. the `LabelMap`s every time you apply an intensity transform, even though it's not affected. I'm not sure what the best approach is.

fepegar:
[I edited your code a bit to understand it more easily, sorry]

romainVala:
sorry for the bad variable name ... (i'll make an effort next time)

Ok then it may not be a bug, since it is a choice, but if you keep it like, that it clearly need to be well documented.
to different behaviour, regarding the copy issue, is very confusing

I can understand the different choice for LabelMap but why then a different behaviour between blur and randomAffine ?

fepegar:
> I can understand the different choice for LabelMap but why then a different behaviour between RandomBlur and RandomAffine ?

[...]

PR #413, opened 2021-01-05
Add option to keep copies of original images

### LINKED ISSUE #93: Add option to keep original image in sample after applying a transform
As discussed in #26.

fepegar:
Resolves #93.

**Description**
This adds a kwarg to keep images in subject untouched after applying the transform.

**Checklist**

### DIFF
--- a/tests/transforms/test_transforms.py
+++ b/tests/transforms/test_transforms.py
@@ -209,6 +209,21 @@ class TestTransforms(TorchioTestCase):
         with self.assertWarns(DeprecationWarning):
             tio.RandomNoise(keys=['t2'])
 
+    def test_keep_original(self):
+        subject = copy.deepcopy(self.sample_subject)
+        old, new = 't1', 't1_original'
+        transformed = tio.RandomAffine(keep={old: new})(subject)
+        assert old in transformed
+        assert new in transformed
+        self.assertTensorEqual(
+            transformed[new].data,
+            subject[old].data,
+        )
+        self.assertTensorNotEqual(
+            transformed[new].data,
+            transformed[old].data,
+        )
+
 
 class TestTransform(TorchioTestCase):
 
--- a/torchio/transforms/data_parser.py
+++ b/torchio/transforms/data_parser.py
@@ -59,8 +59,10 @@ class DataParser:
         elif isinstance(self.data, dict):  # e.g. Eisen or MONAI dicts
             if self.keys is None:
                 message = (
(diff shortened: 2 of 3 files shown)


dmus:
Just a question, does something like this work:

```
transform =tio.Compose([
    tio.Resample(...), 
    tio.Pad(...),
    tio.RescaleIntensity((0, 1))
])
subject_transformed = transform(subject_original)
prediction_data = predict(subject_transformed['mri'][tio.DATA])
prediction = Image(data=prediction_data, affine=subject_transformed.get_first_image()[tio.AFFINE]
inverse_transform = subject_transformed.get_inverse_transform()
subject_original['prediction'] = inverse_transform(prediction)
```

I mean, is subject_original untouched?

romainVala:
Great I haded a similar change in my fork, Just curious what application do you have in mind with this ?

fepegar:
@dmus 

> I mean, is subject_original untouched?

It should be, but there are certain subtleties that need to be taken care of. There's a balance between consistency and efficiency, and I haven't had time to think about this deeply. See #345, for example.

In your example, instead of `Image` I would use `LabelMap` if you want to undo the transform using NN interpolation, and `ScalarImage` if you don't. Also, I would use `subject_original.add_image(inverse_transform(prediction), 'prediction')` for the last line, so you can access it as `subject.prediction`.

---

@romainVala 

> Great I haded a similar change in my fork, Just curious what application do you have in mind with this?

None, I just remembered that you were interested in this and thought that it was now easy to implement. I suppose this is useful for e.g. regression and registration, on which I'm not working. Unfortunately, I'm very biased to brain MRI segmentation!

dmus:
So I guess then I should make it like this to make sure the original data is untouched?:

```
transform =tio.Compose([
    tio.Resample(...), 
    tio.Pad(...),
    tio.RescaleIntensity((0, 1))
])
subject_original_copy = deepcopy(subject_original)
subject_transformed = transform(subject_original_copy)
prediction_data = predict(subject_transformed['mri'][tio.DATA])
prediction = ScalarImage(data=prediction_data, affine=subject_transformed.get_first_image()[tio.AFFINE]
inverse_transform = subject_transformed.get_inverse_transform()
subject_origina.add_image(inverse_transform(prediction))
```

But seeing the unit test this is not what this PR is about?

fepegar:
I don't think you need the `deepcopy`. The transform makes a shallow copy, so you should be fine.

But you're correct, this PR is unrelated!

PR #429, opened 2021-01-26
Run Travis CI on different operating systems

fepegar:


NMontanaBrown:
@fepegar mind if I pick this up?

fepegar:
Please! Do you think it'd make sense to switch to GH Actions directly instead?

NMontanaBrown:
@fepegar should be possible, since GHA has runners with multiple python envs and tox support, as well as configurable deploy actions. What OS support do you want? You have Xenial in your current travis build, but GHA also has Bionic Beaver (18.04) and Focal Fossa (20.04) support. For MacOS they currently only support Catalina (10.15).

fepegar:
I suppose something like `ubuntu-latest` would do. I don't know much about CI/CD, but I guess something like this would be ideal:
- Linux: linting, tests, coverage, deployment
- Windows: tests
- macOS: tests

What do you think?

NMontanaBrown:
Sounds sensible, I'll get on it then 😄

PR #516, opened 2021-04-29
Add functions to process batches

fepegar:
**Description**
When after inference, it's a bit awkward to generate instances of `Subject` from a data loader batch. The new functions in `torchio.utils` will make this much more convenient.

```python
loader = DataLoader(subjects_dataset)  # dataset defined somewhere else
for batch in loader:
    x, y = prepare_batch(batch)  # function defined somewhere else
    labels = model(x).argmax(dim=1, keepdim=True)  # model defined somewhere else
    subjects = tio.utils.get_subjects_from_batch(batch)
    add_images_from_batch(subjects, labels, class_=tio.LabelMap, name='seg')
    for subject in subjects:
        subject.seg.save(subject.seg['filename'])
```

**Checklist**


romainVala:
I do not fully follow your code example, 
what is in batch ?
and prepare_batch ... ? it's a user define function ?

you want to add the 2 functions  get_subjects_from_batch and  add_images_from_batch ?

fepegar:
> I do not fully follow your code example,
what is in batch ?
and prepare_batch ... ? it's a user define function ?
you want to add the 2 functions get_subjects_from_batch and add_images_from_batch ?

Hi, Romain. Thanks for taking a look. `batch` is the item returned by a data loader reading from a subjects dataset, i.e., a dictionary whose keys are the names of the images in the subjects.

`get_subjects_from_batch` is a convenient way of extracting TorchIO-friendly objects from the batch.
`add_images_from_batch` is also convenient to add the predictions to those objects.

Here's a more complete, hopefully self-contained and reproducible example:

```python
import torch
import torchio as tio


def prepare_batch(batch):
    x = batch['image'][tio.DATA]
    y = batch['label'][tio.DATA]
    return x, y


torch.set_grad_enabled(False)

model = torch.nn.Identity()
dataset = tio.datasets.IXITiny('ixi', download=True)
loader = torch.utils.data.DataLoader(dataset, batch_size=4)

for batch in loader:
    x, y = prepare_batch(batch)
    labels = model(x).argmax(dim=1, keepdim=True)
    subjects = tio.utils.get_subjects_from_batch(batch)
    tio.utils.add_images_from_batch(
        subjects,
        labels,
        class_=tio.LabelMap,
        name='seg',
    )
    for subject in subjects:
        filename = subject.seg['filename']
        subject.seg.save(filename.replace('.nii', '_seg.nii'))
    break  # do it for one batch only
```

Hopefully, the docs for these functions will help understand what they do. I should expose the `utils` module in the docs, though.

romainVala:
ok I see
make sense
Thanks

PR #518, opened 2021-05-04
fixed typo in quickstart

DaGuT:
**Description**
Hello,

This is fix for doc.
I've just started using torchio and used quickstart as example to the project I want to start.
I've encountered a error:

```
NameError                                 Traceback (most recent call last)
<ipython-input-5-df67c26faa17> in <module>
     20 # apply RandomAffine 80% of the times and RandomElasticDeformation the rest
     21 # Also, there is a 25% chance that none of them will be applied
---> 22 spatial = OneOf({
     23         tio.RandomAffine(): 0.8,
     24         tio.RandomElasticDeformation(): 0.2,

NameError: name 'OneOf' is not defined
```

As you can see, quickstart guide gives error saying that `OneOf` is not defined, and it's not imported anywhere, so I was like "Hmm... Is it torch, albumentation or torchio implementation?". Then I looked at transofrmation and found that there is an implementation of `OneOf`, so I simply added `tio.` to `OneOf` and it worked.

With this PR I just want it to be fixed for everyone so that they will not have to spend extra time looking for their answer (which `OneOf` is that). Also it'll be helpful for starters to have working solution :)

**Checklist**

### DIFF
--- a/docs/source/quickstart.rst
+++ b/docs/source/quickstart.rst
@@ -71,7 +71,7 @@ in parallel::
     # As RandomAffine is faster then RandomElasticDeformation, we choose to
     # apply RandomAffine 80% of the times and RandomElasticDeformation the rest
     # Also, there is a 25% chance that none of them will be applied
-    spatial = OneOf({
+    spatial = tio.OneOf({
             tio.RandomAffine(): 0.8,
             tio.RandomElasticDeformation(): 0.2,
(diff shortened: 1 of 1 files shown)


fepegar:
Thanks for your contribution, @DaGuT!

@all-contributors please add @DaGuT for doc

PR #527, opened 2021-05-10
Add Mask transform

### LINKED ISSUE #521: Mask values outside of a LabelMap
**🚀 Feature**

Have an IntensityTransform that sets values outside of a certain LabelMap to a specific value.

**Motivation**

This can be handy for example for skull-stripping (as also discussed in #428 ) when a brain mask is already available or in general when a foreground mask is available and you want to mask out the 'background'

**Pitch**

Easily use a transform to mask out certain parts of the image.

**Alternatives**

It is possible to do intensity normalization with a mask using the current IntensityTransforms, and these only take into account the values within the mask, but it doesn't seem to be actually possible to set the values outside of the mask. 

**Additional context**

I can implement this feature but wanted to open a request first according to the contributing guidelines and to check whether there's not something that I missed.

Svdvoort:
Fixes #521.

**Description**

Preprocessing option to set values outside of a LabelMap to a constant value (0 by default).
Also allows to select the specific Labels from the LabelMap that should be considered 'foreground' in case there are multiple Labels.

**Checklist**


Svdvoort:
After some testing I think it's due to the way `get_mask_from_masking_method` returns the mask. I expected it to be the actual values of the `LabelMap`, so in the case of the example [0, 1, 2, 3, 4], but on line 436 it is already turned into boolean. So in this case it returns [False, True, True, True, True]. The specified labels [1, 2] are essentially [True, True], so it only masks the first value of `image_tensor`. 

Is it intended that the returned mask is only boolean? If so what do you propose as an alternate solution to get the `LabelMap`?

fepegar:
> After some testing I think it's due to the way get_mask_from_masking_method returns the mask. I expected it to be the actual values of the LabelMap, so in the case of the example [0, 1, 2, 3, 4], but on line 436 it is already turned into boolean. So in this case it returns [False, True, True, True, True]. The specified labels [1, 2] are essentially [True, True], so it only masks the first value of image_tensor.

You are right, thanks for investigating.

> Is it intended that the returned mask is only boolean?

I think it makes sense that `get_mask_from_masking_method` returns a binary image, yes. What about adding a `labels` kwarg to that method and compute the mask using the labels if they are not `None`?

Svdvoort:
> I think it makes sense that `get_mask_from_masking_method` returns a binary image, yes. What about adding a `labels` kwarg to that method and compute the mask using the labels if they are not `None`?

With this `labels` argument, it would still return a binary image? So give it labels=[1, 3, 4] would return a binary mask for only those labels? It would solve this issue, but I can imagine that for other use cases it might also be handy to have the actual mask (with the labels, not as a binary image) returned. We could also add a `return_labels` argument which is False by default, but if it is True it would return the mask with the labels instead of a binary image? 

However, if the underlying assumption is always that the function should return a binary image then I agree with the first approach.

fepegar:
I think it's safe to assume that a binary image is returned. We can modify the behavior later on, if new use cases are needed.

Svdvoort:
I've added an option to return a binary mask from only the requested labels, and updated this in the `mask` function, it now passes the tests.

fepegar:
Thanks! I'll fix the conflict and merge ASAP.

fepegar:
Thanks for your contribution, @Svdvoort!

@all-contributors please add @Svdvoort for code

PR #542, opened 2021-05-20
Fix unexpected exception handling

### LINKED ISSUE #540: Unexpected exception handling
**🐛Bug**

First, thanks for this awesome project!
There is an unexpected exception handling when parsing the image path.

https://github.com/fepegar/torchio/blob/bf0a86fa5832b7236932a6e0cdd0c19c8a6ce724/torchio/data/image.py#L380-L383

I think the implementation here may be better if it is in the following way. Can I send a PR for this?
Is there any other concern? :)

```python
if isinstance(path, Iterable):
    return [self._parse_single_path(p) for p in path]
return self._parse_single_path(path)
```

**To reproduce**

```python
import torchio as tio

tio.Image(0)
```

**Expected behavior**

```python-traceback
TypeError: expected str, bytes or os.PathLike object, not int
```

**Actual behavior**

```python-traceback
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/siahuat0727/.local/lib/python3.8/site-packages/torchio/data/image.py", line 141, in __init__
    self.path = self._parse_path(path)
  File "/home/siahuat0727/.local/lib/python3.8/site-packages/torchio/data/image.py", line 383, in _parse_path
    return [self._parse_single_path(p) for p in path]
TypeError: 'int' object is not iterable
```

siahuat0727:
Fixes #540.

**Description**
Better exception handling

**Checklist**

### DIFF
--- a/torchio/data/image.py
+++ b/torchio/data/image.py
@@ -1,5 +1,6 @@
 import warnings
 from pathlib import Path
+from collections.abc import Iterable
 from typing import Any, Dict, Tuple, Optional, Union, Sequence, List, Callable
 
 import torch
@@ -377,10 +378,10 @@ class Image(dict):
             ) -> Optional[Union[Path, List[Path]]]:
         if path is None:
             return None
-        if isinstance(path, (str, Path)):
-            return self._parse_single_path(path)
-        else:
+        if isinstance(path, list):
             return [self._parse_single_path(p) for p in path]
+        else:
+            return self._parse_single_path(path)
 
     def _parse_tensor(
             self,
--- a/torchio/data/image.py
+++ b/torchio/data/image.py
@@ -378,7 +378,7 @@ class Image(dict):
             ) -> Optional[Union[Path, List[Path]]]:
         if path is None:
             return None
-        if isinstance(path, list):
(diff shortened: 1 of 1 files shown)


torchio/data/image.py
```diff
@@ -377,10 +378,10 @@ def _parse_path(
             ) -> Optional[Union[Path, List[Path]]]:
         if path is None:
             return None
-        if isinstance(path, (str, Path)):
-            return self._parse_single_path(path)
-        else:
+        if isinstance(path, list):
```
fepegar, line 381: Did you mean `Iterable` here?

fepegar (changes requested):

torchio/data/image.py
siahuat0727: Ops... I changed it at first to try something else and then forgot about it.
Can I use `git commit --amend` and then force push for this situation?

fepegar: You can just fix this in a new commit, don't worry about squashing or force-pushing!

fepegar:
Thanks again, @siahuat0727.

@all-contributors please add @siahuat0727 for bug

PR #584, opened 2021-06-23
Discussion #489: add CopyAffine transform

Albans98:
Fixes #489 .

**Description**
Add a CopyAffine transform with a target_key parameter to copy the affine from this key to all LabelMap objects of a given torchio subject.

**Checklist**

### DIFF
--- a/torchio/transforms/preprocessing/spatial/copy_affine.py
+++ b/torchio/transforms/preprocessing/spatial/copy_affine.py
@@ -1,59 +1,79 @@
-from ... import SpatialTransform
-from ....data.subject import Subject
-import numpy as np
 import copy
+from ....data.subject import Subject
+from ... import SpatialTransform
 
 
 class CopyAffine(SpatialTransform):
-    """
-    Copy the affine of a target image to all images that have a
-    different affine than the target.
+    """Copy the spatial metadata from a reference image in the subject.
 
-    This transform should be used with caution: be sure that you know
-    what you are doing.
-    In general having different affines between the volumes
-    need to be taken into account with resampling or cropping transform.
-    For some special cases related to the rounding error in the affine nifti
-    parameter or because the affine has been changed (for instance after a
-    spm coregister) one may want to erase the original affine read from the
-    nifti volume and take the one of another volume.
-    But you must check if it is correct to copy the reference affine.
-    This is fine if you have the same voxel grid and the same voxel ordering.
+    Small unexpected differences in spatial metadata across different images
+    of a subject can arise due to rounding errors while converting formats.
+
+    If the ``shape`` and ``orientation`` of the images are the same and their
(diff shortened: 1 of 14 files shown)


romainVala:
merci alban

torchio/transforms/preprocessing/spatial/copy_affine.py
```diff
@@ -0,0 +1,31 @@
...
+                 target_key: str,
+                 **kwargs):
+        super().__init__(**kwargs)
+        self.target_key = target_key
+        self.args_names = ['target_key']
+
+    def apply_transform(self, subject: Subject) -> Subject:
+        affine = subject[self.target_key].affine
+        for image in subject.get_images(intensity_only=False):
+            if isinstance(image, LabelMap):
```
fepegar, line 29: Why does this affect label maps only?

fepegar (changes requested):
Thanks for your PR! I added a comment. Could you please update the corresponding rst file in the documentation?

torchio/transforms/preprocessing/spatial/copy_affine.py
romainVala: I agree, it should be all, except the target_key

romainVala:
Hi, I am not sure to understand what the error is, 
should we change things ... ?

fepegar:
> Hi, I am not sure to understand what the error is,
should we change things ... ?

I think the error is happening because of a Sphinx problem that was fixed in 6e614478ed3b273613c7a08ca2b3f5ed2bef6c52, before this branch was created. Don't worry about it. I'm going to take a look at this PR tonight or tomorrow.

fepegar:
Thanks for your contribution, @Albans98!

@all-contributors please add @Albans98 for code

Albans98:
Thanks for your time and your reviews !

romainVala:
@Albans98 just find out, there is one line missing at the end of the init
```python
        self.args_names = ['target']
```

fepegar:
That's my fault, I removed it thinking it was necessary. It's needed to reproduce the transform, right?

romainVala:
may be @Albans98  can confirm, I don't know what the exact problem was, 
In our use case, we do save the transform history (to be able to reproduce later)

but anyway It make sens to have it for all the transform (preprocessing and augmentation)
(I quickly check, and there may be some other missing like OneHot)

Albans98:
Hello, yes the problem is when we try to save the history of all applied transforms, we get an error 'TypeError: __init__() missing 1 required positional argument: 'target''. 

Adding the target in the list of args makes it possible to trace back to the 'target' value.

fepegar:
Ok, let me fix that.

fepegar:
Done. Thanks for pointing that out!

Albans98:
Thank you for fixing this !

PR #592, opened 2021-07-01
Add tolerance for numeric checks

justusschock:
**Description**
This PR combines the previous approaches in checking for consistent affines and other attributes to add tolerances for numeric values and a fallback for other types.

I am not sure on whether to name this a breaking change since this introduces tolerances and also changes the error messages slightly. However, the overall behaviour would stay the same.

**Checklist**

### DIFF
--- a/torchio/data/subject.py
+++ b/torchio/data/subject.py
@@ -214,19 +214,45 @@ class Subject(dict):
 
     def clear_history(self) -> None:
         self.applied_transforms = []
-
-    def check_consistent_attribute(self, attribute: str) -> None:
-        values_dict = {}
+        
+    def check_consistent_attribute(self,
+                               attribute: str,
+                               rtol=1e-6,
+                               atol=1e-6,
+                               message: Optional[str] = None) -> None:
         iterable = self.get_images_dict(intensity_only=False).items()
-        for image_name, image in iterable:
-            values_dict[image_name] = getattr(image, attribute)
-        num_unique_values = len(set(values_dict.values()))
-        if num_unique_values > 1:
+        if message is None:
             message = (
-                f'More than one {attribute} found in subject images:'
-                f'\n{pprint.pformat(values_dict)}'
-            )
-            raise RuntimeError(message)
+                f'More than one value for {attribute} found in subject images:'
+                '\n{}')
+        try:
+            first_attr = None
+            first_image = None
+
+            for image_name, image in iterable:
+                if first_attr is None:
+                    first_attr = getattr(image, attribute)
+                    first_image = image_name
+
+                else:
+                    curr_attr = getattr(image, attribute)
+                    if not np.allclose(curr_attr, first_attr, rtol=rtol,
+                                       atol=atol):
+                        message = message.format(
+                            pprint.pformat({
+                                first_image: first_attr,
+                                image_name: curr_attr
+                            }))
+
+        except TypeError:
+            # fallback for non-numeric values
+            values_dict = {}
+            for image_name, image in iterable:
+                values_dict[image_name] = getattr(image, attribute)
+            num_unique_values = len(set(values_dict.values()))
+            if num_unique_values > 1:
+                message = message.format(pprint.pformat(values_dict))
+                raise RuntimeError(message)
 
     def check_consistent_spatial_shape(self) -> None:
         self.check_consistent_attribute('spatial_shape')
@@ -235,24 +261,7 @@ class Subject(dict):
         self.check_consistent_attribute('orientation')
 
     def check_consistent_affine(self):
-        # https://github.com/fepegar/torchio/issues/354
-        affine = None
-        first_image = None
-        iterable = self.get_images_dict(intensity_only=False).items()
-        for image_name, image in iterable:
-            if affine is None:
-                affine = image.affine
-                first_image = image_name
-            elif not np.allclose(affine, image.affine, rtol=1e-6, atol=1e-6):
-                message = (
-                    f'Images "{first_image}" and "{image_name}" do not occupy'
-                    ' the same physical space.'
-                    f'\nAffine of "{first_image}":'
-                    f'\n{pprint.pformat(affine)}'
-                    f'\nAffine of "{image_name}":'
-                    f'\n{pprint.pformat(image.affine)}'
-                )
-                raise RuntimeError(message)
+        self.check_consistent_attribute('affine')
 
     def check_consistent_space(self):
         self.check_consistent_spatial_shape()


fepegar:
Hi, @justusschock. I thought the plan was to use the new transform from #584, on which @Albans98 is working.

justusschock:
Hi @fepegar ,

This is not for copying, but the issue I am experiencing is that when reading files from different datatypes (i.e. my images are stored in dicom and my labels in nifti) different readers produce slightly differen spacings like this:
```
'data': (0.52734375, 0.52734375, 1.5999999999999999),
'label': (0.52734375, 0.52734375, 1.600000023841858)
```

When comparing them with some tolerance, these are identical, but due to this issue, I cannot use any of the subject level properties.

I think the transform from @Albans98 would come handy, when the labels don't contain information about the affine at all (e.g by just being tensors/arrays/images from formats without metadata support. So I think those are two different situations.

fepegar:
I guess my question is whether you can't use the new transform to bypass this issue, e.g. you can add `CopyAffine('data')` before the operation that raises the error about the different spatial metadata being found. What do you think?

justusschock:
I thought about that, but unfortunately I don't think that's suitable in my case, since I need to get those properties before actually instantiating the transforms. Also I don't want the check to be neglected since IMO it's an important sanity check. I just don't want it to allow for really minor offs due to numeric issues.

fepegar:
That's ok, I just wanted to understand. I think that makes sense. I will take a look as soon as I can!

justusschock:
Thanks :)

romainVala:
Just be aware that to give some more tolerance on the check, may lead to some difference when applying transform 
I thing we add a 1e-10 difference in voxel size, but then using a resampling, will lead for the transformed data to one voxel more in the grid ... which is not possible to work with ...

That is why we need the CopyAffine #584 

I do not understand 

> I think the transform from @Albans98 would come handy, when the labels don't contain information about the affine at all (e.g by just being tensors/arrays/images from formats without metadata support). So I think those are two different situations.

when you create an image from a tensor, you have to specify an affine , so it will be alway present .. no ?

justusschock:
@romainVala why is the tolerance so low? AFAIK for 32 bits anything beyond 1e-7 or maybe 1e-8 can come from numeric instabilities.

I'm fine with reducing the tolerance, just having some would already be great (and maybe an option to specify this).

> when you create an image from a tensor, you have to specify an affine , so it will be alway present .. no ?

You should but if you don't you end up with a default identity affine.

romainVala:
I do agree that the both affine are equal but still ...
My concern is that even with very low difference, you can end up with inconsistency you do not want.

Here is an example 
```python
import numpy as np
import torch
import torchio as tio

img = torch.randn(1, 512, 512, 100)
mask = torch.tensor(img > 0).type(torch.int16)

af1 = np.array([[0.8, 0, 0, 0],
                [0, 0.8, 0, 0],
                [0, 0, 2.50000000000001, 0],
                [0, 0, 0, 1]])

af2 = np.array([[0.8, 0, 0, 0],
                [0, 0.8, 0, 0],
                [0, 0, 2.49999999999999, 0],
                [0, 0, 0, 1]])

sub = tio.Subject(
  image = tio.ScalarImage(tensor = img, affine = af1),
  mask = tio.LabelMap(tensor = mask, affine = af2)
)

af1[2,2] - af2[2,2]

```
`Out[32]: 2.042810365310288e-14`
that is pretty small, but now imagine you want to resample to a given voxel resolution 2.5

```python
t=tio.Resample(2.5)
r_sub = t(sub)

r_sub.image.data.shape
Out[39]: torch.Size([1, 164, 164, 101])
r_sub.mask.data.shape
Out[40]: torch.Size([1, 164, 164, 100])

```

Resample is probably the only transform that will lead to discrepancy in the finale images shape. and we can manage it with an additional crop or pad,

torchio/data/subject.py
```diff
@@ -214,19 +214,46 @@ def apply_inverse_transform(self, **kwargs) -> 'Subject':
 
     def clear_history(self) -> None:
         self.applied_transforms = []
-
-    def check_consistent_attribute(self, attribute: str) -> None:
-        values_dict = {}
+        
+    def check_consistent_attribute(self,
```
fepegar, line 218: The indentation should be consistent with the other methods in the class.

torchio/data/subject.py
```diff
@@ -214,19 +214,46 @@ def apply_inverse_transform(self, **kwargs) -> 'Subject':
...
+                               attribute: str,
+                               rtol=1e-6,
```
fepegar, line 220: There should be typing hints (`float`) for the tolerance arguments.

torchio/data/subject.py
```diff
@@ -214,19 +214,46 @@ def apply_inverse_transform(self, **kwargs) -> 'Subject':
...
+                               atol=1e-6,
```
fepegar, line 221: Can we rename `rtol` and `atol` to something more descriptive? E.g., `relative_tolerance` and `absolute_tolerance` or, if that's too long, `rel_tolerance` and `abs_tolerance`.

It would be also nice to have a little docstring explaining what these are, with at least one example.

torchio/data/subject.py
```diff
@@ -214,19 +214,46 @@ def apply_inverse_transform(self, **kwargs) -> 'Subject':
...
+                               message: Optional[str] = None) -> None:
         iterable = self.get_images_dict(intensity_only=False).items()
-        for image_name, image in iterable:
-            values_dict[image_name] = getattr(image, attribute)
-        num_unique_values = len(set(values_dict.values()))
-        if num_unique_values > 1:
+        if message is None:
```
fepegar, line 224: I like this message. Why not just hard-coding it? I'm not sure the user needs the option to have a custom message. They can always catch the exception and print something else themselves.

torchio/data/subject.py
```diff
@@ -214,19 +214,46 @@ def apply_inverse_transform(self, **kwargs) -> 'Subject':
...
+                    if not np.allclose(curr_attr, first_attr, rtol=rtol,
+                                       atol=atol):
+                        message = message.format(
+                            pprint.pformat({
+                                first_image: first_attr,
+                                image_name: curr_attr
+                            }))
+                        raise RuntimeError(message)
+
+        except TypeError:
```
fepegar, line 248: This is currently not covered by the tests. When does this happen?

fepegar (changes requested):
Hi, @justusschock. I added some comments.

Something else. Would these changes be meaningful for your use case? Wouldn't you also need to be able to specify the tolerance? I thought you wanted to be able to specify it when instantiating the subject.

torchio/data/subject.py
justusschock: ll add that :)

torchio/data/subject.py
justusschock: Will add that :)

torchio/data/subject.py
justusschock: I didn't think about the user, rather when this method will be used for other checks so that the current methods these checks run on, can be reused and don't introduce breaking changes :)

torchio/data/subject.py
justusschock: Whenever something cannot be converted to a numpy numeric value. I guess that was on me, when I wrote a custom transform that had to check the orientation beforehand for example. And this method is now sufficiently generic, that it also allows that :)

justusschock:
Hi @fepegar , sorry for getting back on this so late, Been quite busy with some other stuff.

> Would these changes be meaningful for your use case? Wouldn't you also need to be able to specify the tolerance? I thought you wanted to be able to specify it when instantiating the subject.

Yes they would. The standard tolerance would be sufficient for me. Alternatively, we could also think about adding them as optional arguments to the init. Not sure how much you like that/are open to it?

PR #609, opened 2021-07-21
Extend the queue to take different samples per volume for each subject in the dataset

jmlipman:
Implements idea described in Discussion #604.

**Description**
This modification permits passing a list of ints to Queue's samples_per_volumes. As discussed [here](https://github.com/fepegar/torchio/discussions/604), it can be useful to give a different sample_per_volume to each subject when dealing with 2D slices of 3D images that have a very different stack size.

**Checklist**


torchio/data/queue.py
```diff
@@ -146,15 +147,39 @@ def __init__(
         self.shuffle_subjects = shuffle_subjects
         self.shuffle_patches = shuffle_patches
         self.samples_per_volume = samples_per_volume
+        if type(samples_per_volume) == int:
```
fepegar, line 150: Please use `isinstance`, as in https://github.com/fepegar/torchio/blob/a51248ee52382bc639f45e46c98032a5ed33f33b/torchio/data/subject.py#L69

torchio/data/queue.py
```diff
@@ -146,15 +147,39 @@ def __init__(
...
+            self.samples_per_volume = [samples_per_volume for _ in range(
```
fepegar, line 151: You could just do `self.num_subjects * [samples_per_volume]`.

torchio/data/queue.py
```diff
@@ -146,15 +147,39 @@ def __init__(
...
+                self.num_subjects)]
+        assert type(self.samples_per_volume) == list, ('`samples_per_volume`'
```
fepegar, line 153: Please relax this to `Sequence` (with `isinstance`) and raise a `ValueError`, instead of using `assert`.

torchio/data/queue.py
```diff
@@ -146,15 +147,39 @@ def __init__(
...
+                                                       'should be an int'
+                                                       'or a list')
+        assert len(self.samples_per_volume) == self.num_subjects, (
```
fepegar, line 156: Please use `ValueError` here as well.

torchio/data/queue.py
```diff
@@ -146,15 +147,39 @@ def __init__(
...
         self.verbose = verbose
         self._subjects_iterable = None
-        if start_background:
-            self._initialize_subjects_iterable()
         self.patches_list: List[Subject] = []
         self.num_sampled_patches = 0
 
+        # Same random shuffling applied to subjects and volumes
+        if self.shuffle_subjects:
+            tmp_seed = np.random.random()
```
fepegar, line 168: Is it possible to use PyTorch instead of NumPy for this? NumPy random operations are sometimes problematic in this context.

torchio/data/queue.py
```diff
@@ -146,15 +147,39 @@ def __init__(
...
+            random.Random(tmp_seed).shuffle(self.subjects_dataset._subjects)
```
fepegar, line 169: Maybe it's not a good idea to modify the dataset instance. Do you know if there's a way to have a local, shuffled version of the dataset? Maybe using PyTorch `RandomSampler`?

torchio/data/sampler/uniform.py
```diff
@@ -24,7 +24,7 @@ def _generate_patches(
         patches_left = num_patches if num_patches is not None else True
         while patches_left:
             index_ini = [
-                torch.randint(x + 1, (1,)).item()
+                torch.randint(int(x) + 1, (1,)).item()
```
fepegar, line 27: Why was this necessary?

fepegar (changes requested):
Thanks for the PR. I have added some comments. I will verify the new behavior once the comments have been addressed. Sorry for the delayed responses, I'm currently on holiday!

torchio/data/sampler/uniform.py
jmlipman: I guess this has never been a problem, so I removed it. But in my laptop (Python 3.8.9, Pytorch 1.4.0) this raises an error: `torch.randint(np.int64(5) + 1, (1,))` so I had to cast it to int. In my other computer (newer version of Pytorch) it doesn't raise an error.

torchio/data/queue.py
jmlipman: I can do this. Why not raising TypeError?

fepegar: You're right, `TypeError` would be more appropriate.

torchio/data/sampler/uniform.py
fepegar: Why not just leave it as it was? If it's an issue, it should be addressed separately as it is out of the scope of this PR.

torchio/data/queue.py
jmlipman: Like passing a cloned (shuffled) `self.subjects_dataset` to the DataLoader in `_get_subjects_iterable()` while not touching self.subjects_dataset?

jmlipman: I've just sent a new pull request tackling the other comments and trying to do this.

fepegar:
Just added some minor edits. I didn't expect so much new code. Can you please summarize the new behavior?

torchio/data/queue.py
```diff
@@ -250,14 +296,24 @@ def _get_first_item(batch):
     def _get_subjects_iterable(self) -> Iterator:
         # I need a DataLoader to handle parallelism
         # But this loader is always expected to yield single subject samples
+
+        # Same random shuffling applied to subjects and volumes
+        if self.shuffle_subjects:
+            random_idx = list(RandomSampler(self.subjects_dataset))
```
fepegar, line 302: Have you tried this? I might be wrong, but I suspect this will load all the data in the dataset from disk, which is really undesirable.

jmlipman: I haven't tested RandomSampler with real data, but by looking at the [https://github.com/pytorch/pytorch/blob/master/torch/utils/data/sampler.py#L113](code) it seems that it doesn't load the data; it works directly with len(data). In the end, random_idx becomes a randomized list of indices

jmlipman:
Hola Fernando,

Feel free to leave this until you are back from your vacation; I will still continue pulling this contribution : ) Also, do you know how can I pull your edits to my local machine? I'm not an expert in git and `git pull` (alone) won't work since it's pulling from my forked version of torchio (it returns me "There is no tracking information for the current branch. Please specify which branch you want to merge with.")

> Just added some minor edits. I didn't expect so much new code. Can you please summarize the new behavior?

I leave a brief and detailed explanation. Let me know if I can further clarify my code or if you have any other suggestion.

Brief explanation:
Now we need to keep track of how many samples per volume we are going to extract, and since they might not be the same for each volume, we can use a list of counters, which is a copy of the `samples_per_volume` that the user passed. Inside fill, we decrease these counters as we fill `patches_list`, and we control from which subject we will extract patches.

Detailed explanation:
Constructor:
The main idea is to use the list of `samples_per_volume` to keep track of the different samples per volume that we need to extract. For this reason, I clone it (`self.counter_samples_per_volume`) and use this clone as a counter (more precisely, a list of counters per volume). Additionally, we keep track of `curr_subject` ("current subject" from which we are extracting the patches) and `idx_subject` (its position in the list of counters) in memory.

Fill:
We enter the first if at the end of the epoch: since we are decreasing the list of counters `counter_samples_per_volume`, we need to fill it back. Same thing with `idx_subject` and `curr_subject`, which obtain their initial values (see Constructor).

Inside the while, as before this PR, we fill patches_list. The stopping criteria of the while are the following: 1) if the number of patches inside `patches_list` is more than the allowed, 2) or there are no more patches to put inside `patches_list`, 3) or there are no more subjects to extract patches.

If the counter in the current subject position is zero, we move to the next subject for extracting patches. Now, there are two possible situations, and the variable `spv` (stands for "samples per volume") will temporarily store how many patches from the current subject we can extract.

Case A: the number of patches we want to extract from the current subject (plus the number of already extracted patches in `self.patches_list`) is more than the limit (`self.max_length`). In this case, we can only take a portion of patches of the current subject. Example: say want to extract 40 patches from the current subject, and we already have 20 in patches_list (from the previous subject), but our `self.max_length` is 50. Then, we can only extract 30 patches of the current subject, and in the next fill, we will extract the remaining 10.

Case B (¬A): If A) is not the case, i.e., in the previous example, say that our `patches_list` is empty; then, we can fill `patches_list` with all the 40 patches of current subject.

Finally, we randomize `patches_list` if necessary; otherwise, we reverse it so that pop() extracts the patches in order.

jmlipman:
Hi Fernando! No worries. I'm glad that you are considering this idea : )

fepegar:
Closing in favor of #795.

@all-contributors please add @jmlipman for ideas and code

PR #348, opened 2020-11-04
347 clipping

sarthakpati:
Fixes #347.

**Description**
Adds a transformation that performs thresholding of scalar images

**Checklist**

- [ ] I have read the [`CONTRIBUTING`](https://github.com/fepegar/torchio/blob/master/CONTRIBUTING.rst) docs and have a developer setup (especially important are `pre-commit`and `pytest`)
- [x] Non-breaking change (would not break existing functionality)
- [ ] Breaking change (would cause existing functionality to change)
- [x] Tests added or modified to cover the changes
- [ ] Integration tests passed locally by running `pytest`
- [x] In-line docstrings updated
- [ ] Documentation updated, tested running `make html` inside the `docs/` folder
- [x] This pull request is ready to be reviewed
- [ ] If the PR is ready and there are multiple commits, I have [squashed them and force-pushed](https://www.w3docs.com/snippets/git/how-to-combine-multiple-commits-into-one-with-3-steps.html#force-pushing-commits-7)

PR #622, opened 2021-08-16
Added intensity clamp transform

### LINKED ISSUE #347: A new transform that would clip intensities
**🚀 Feature**

An "intensity clipping" transform would be very useful.

**Motivation**

When working with CT images, where the intensities have physical meaning, for example, to pick up liver in abdominal CT, we can simply pick up all intensities between `[-50,200]` and make the rest zero.

**Pitch**

Basing on [RescaleIntensity](https://torchio.readthedocs.io/transforms/preprocessing.html?highlight=RescaleIntensity#torchio.transforms.RescaleIntensity), a **ClipIntensity** transform could be implemented, which takes hard intensity ranges and not percentiles as input.

**Alternatives**

It would be feasible to put this into the training loop directly, but that kind of defeats the purpose.

**Additional context**

N.A.

mattwarkentin:
Fixes #347.

**Description**
Added `torchio.ClampIntensity()` transform according to discussions in #347 and #348. I chose to use the word "clamp" to be consistent with `torch.clamp` terminology (which is also the underlying implementation), but this can be changed according to your preference (e.g. clipping).

Added tests and all tests passed locally.

**Checklist**

### DIFF
--- a/torchio/transforms/preprocessing/intensity/clamp.py
+++ b/torchio/transforms/preprocessing/intensity/clamp.py
@@ -1,20 +1,21 @@
+from torchio.data.image import ScalarImage
 from typing import Optional
 import torch
 from ....data.subject import Subject
-from .normalization_transform import NormalizationTransform
+from ...intensity_transform import IntensityTransform
 
 
-class ClampIntensity(NormalizationTransform):
+class Clamp(IntensityTransform):
     """Clamp intensity values to a certain range.
     Args:
-        out_min: See :func:`~torch.clamp`.
-        out_max: See :func:`~torch.clamp`.
+        out_min: See :func:`torch.clamp`.
+        out_max: See :func:`torch.clamp`.
 
     Example:
         >>> import torchio as tio
         >>> ct = tio.ScalarImage('ct_scan.nii.gz')
         >>> ct_air, ct_bone = -1000, 1000
-        >>> clamp = tio.ClampIntensity(out_min=ct_air, out_max=ct_bone)
+        >>> clamp = tio.Clamp(out_min=ct_air, out_max=ct_bone)
         >>> ct_clamped = clamp(ct)
     """
     def __init__(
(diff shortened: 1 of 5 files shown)


torchio/transforms/preprocessing/intensity/clamp.py
```diff
@@ -0,0 +1,43 @@
...
+            image_name: str,
+            mask: torch.Tensor,
+            ) -> None:
+        image = subject[image_name]
+        image.set_data(self.clamp(image.data, image_name))
+
+    def clamp(
+            self,
+            tensor: torch.Tensor,
+            image_name: str,
```
fepegar, line 41: Why do you need `image_name` here?

torchio/transforms/preprocessing/intensity/clamp.py
```diff
@@ -0,0 +1,43 @@
+from typing import Optional
+import torch
+from ....data.subject import Subject
+from .normalization_transform import NormalizationTransform
+
+
+class ClampIntensity(NormalizationTransform):
+    """Clamp intensity values to a certain range.
+    Args:
+        out_min: See :func:`~torch.clamp`.
```
fepegar, line 10: I think we don't need the tildes here, so we can see `torch`.

fepegar (changes requested):
Looking good. I've added some comments. I think I messed up and some comments went to the commit in your fork, instead of this PR. Can you see them?

mattwarkentin:
Yep I can see all the comments. Will make these changes. Much of the unnecessary code is probably because I used `RescaleIntensity` as a template and so it needs refactoring now.

torchio/transforms/preprocessing/intensity/clamp.py
```diff
@@ -27,6 +27,13 @@ def __init__(
         super().__init__(**kwargs)
         self.out_min, self.out_max = out_min, out_max
 
+    def __repr__(self):
```
fepegar, line 30: The parent class already has this method. And it would seem a bit arbitrary to have only this one transform have a custom `repr`.

fepegar (changes requested):

torchio/transforms/preprocessing/intensity/clamp.py
mattwarkentin: Should I add the `args_names` attribute to the `Clamp` class so the `repr` looks like how I had it previously?

fepegar: I just did that, thanks. The main importance of `args_names` (which is not well documented and has a bad variable name) is to be able to reproduce the transform applied to a subject, so it was indeed necessary. Good catch!

fepegar:
Are you happy with my changes? Shall I merge?

mattwarkentin:
LGTM! That's kind of you to check with me.

fepegar:
Kind of you to contribute to an open-source project! Thanks again!

PR #625, opened 2021-08-18
Add transform for sliding intensity projections

mattwarkentin:
**Description**

Hi @fepegar,

You may not be interested in this pull request, but this was a feature I was playing around with in my own project and though it might be worth spinning into a proper transform for `torchio`. Intensity projections can perhaps be useful to reduce the dimensionality of a high-resolution image while maintaining aspects of 3D structures that span across axes.

This features provides a transform to do intensity projections across a chosen `axis`, possibly using thin/thick slab sliding projections. The default arguments perform a maximum intensity projection (MIP) across the whole chosen `axis` dimension. The user can configure the `projection_type`, which is `'max'` by default, but supports `'min'`, `'mean'`, `'median'`, and `'quantile'` projections (more could be supported if desired).

In addition, the user can configure the `slab_thickness` and `stride` to perform thin/thick slab sliding projections. Since it may be unknown/variable what the `axis` extent is for all images being projected, the `full_slabs_only` argument controls what to do when the combination of `slab_thickness` and `stride` extends beyond the extent of the image.

I look forward to hearing your thoughts about whether or not this fits into the scope of `torchio`. If so, i am happy to make any changes to improve this PR.

**Checklist**

### DIFF
--- a/torchio/transforms/preprocessing/intensity/projection.py
+++ b/torchio/transforms/preprocessing/intensity/projection.py
@@ -46,7 +46,7 @@ class Projection(IntensityTransform):
         sub = tio.datasets.Slicer('CTChest')
         ct = sub.CT_chest
         axial_mips = tio.Projection("S", slab_thickness=20, stride=20)
-        ct_mips= axial_mips(ct)
+        ct_mips = axial_mips(ct)
         sub.add_image(ct_mips, 'MIP')
         sub.plot()
 
--- a/torchio/transforms/preprocessing/intensity/projection.py
+++ b/torchio/transforms/preprocessing/intensity/projection.py
@@ -86,11 +86,11 @@ class Projection(IntensityTransform):
     def validate_quantile(self):
         message = (
             'For `projection_type="quantile"`, `q` must be a scalar value'
-            f'between 0 and 1, not {self.q}.'
+            f'in the range [0, 1], not {self.q}.'
             )
         if self.q is None:
             raise ValueError(message)
-        elif 0 < self.q < 1:
+        elif 0 <= self.q <= 1:
             pass
         else:
             raise ValueError(message)
@@ -110,6 +110,9 @@ class Projection(IntensityTransform):
         image.set_data(self.projection(image.data))
 
     def projection(self, tensor: torch.Tensor) -> torch.Tensor:
+        if self.projection_type in ['mean', 'quantile']:
+            tensor = tensor.to(torch.float)
+
         if self.full_slabs_only:
             start_index = 0
             num_slabs = 0
--- a/torchio/transforms/preprocessing/intensity/projection.py
+++ b/torchio/transforms/preprocessing/intensity/projection.py
@@ -65,9 +65,9 @@ class Projection(IntensityTransform):
 
     def get_projection_function(self):
         if self.projection_type == 'max':
-            projection_fun = torch.max
+            projection_fun = torch.amax
         elif self.projection_type == 'min':
-            projection_fun = torch.min
+            projection_fun = torch.amin
         elif self.projection_type == 'mean':
             projection_fun = torch.mean
         elif self.projection_type == 'median':
@@ -126,14 +126,14 @@ class Projection(IntensityTransform):
         for _ in range(num_slabs):
             slab_indices = torch.tensor(list(range(start_index, end_index)))
             slab = tensor.index_select(self.axis_index, slab_indices)
-            if self.projection_type == 'mean':
-                projected = self.projection_fun(
+            if self.projection_type == 'median':
+                projected, _ = self.projection_fun(
                     slab, dim=self.axis_index, keepdim=True)
-            elif self.projection_type == 'quantile':
+            if self.projection_type == 'quantile':
                 projected = self.projection_fun(
                     slab, q=self.q, dim=self.axis_index, keepdim=True)
             else:
-                projected, _ = self.projection_fun(
+                projected = self.projection_fun(
                     slab, dim=self.axis_index, keepdim=True)
             slabs.append(projected)
             start_index += self.stride
--- a/torchio/transforms/preprocessing/intensity/projection.py
+++ b/torchio/transforms/preprocessing/intensity/projection.py
@@ -129,7 +129,7 @@ class Projection(IntensityTransform):
             if self.projection_type == 'median':
                 projected, _ = self.projection_fun(
                     slab, dim=self.axis_index, keepdim=True)
-            if self.projection_type == 'quantile':
+            elif self.projection_type == 'quantile':
                 projected = self.projection_fun(
                     slab, q=self.q, dim=self.axis_index, keepdim=True)
             else:
--- a/torchio/transforms/preprocessing/intensity/projection.py
+++ b/torchio/transforms/preprocessing/intensity/projection.py
@@ -35,10 +35,21 @@ class Projection(IntensityTransform):
 
     Example:
         >>> import torchio as tio
-        >>> sub = tio.datasets.Colin27()
-        >>> axial_mips = tio.Projection("S", slab_thickness=20)
-        >>> sub_t = axial_mips(sub)
-        >>> sub_t.t1.plot()
+        >>> ct = tio.datasets.Slicer('CTChest').CT_chest
+        >>> axial_mips = tio.Projection("S", slab_thickness=20, stride=20)
+        >>> ct_t = axial_mips(ct)
+        >>> ct_t.plot()
+
+    .. plot::
+
+        import torchio as tio
+        sub = tio.datasets.Slicer('CTChest')
+        ct = sub.CT_chest
+        axial_mips = tio.Projection("S", slab_thickness=20, stride=20)
+        ct_mips= axial_mips(ct)
+        sub.add_image(ct_mips, 'MIP')
+        sub.plot()
+
     """
     def __init__(
             self,
--- a/torchio/transforms/preprocessing/intensity/projection.py
+++ b/torchio/transforms/preprocessing/intensity/projection.py
@@ -1,0 +1,144 @@
+import torch
+from torchio.data.image import ScalarImage
+from ....data.subject import Subject
+from ...intensity_transform import IntensityTransform
+from typing import Optional
+from math import ceil
+
+
+class Projection(IntensityTransform):
+    """Project intensities along a given axis, possibly with sliding slabs.
+
+    Args:
+        axis: Possible inputs are ``'Left'``, ``'Right'``, ``'Anterior'``,
+                ``'Posterior'``, ``'Inferior'``, ``'Superior'``. Lower-case
+                versions and first letters are also valid, as only the first
+                letter will be used.
+        slab_thickness: Thickness of slab projections. In other words, the
+            number of voxels in the ``axis`` dimension to project across.
+            If ``None``, the projection will be done across the entire span of
+            the ``axis`` dimension (i.e. ``axis`` dimension will be reduced to
+            1).
+        stride: Number of voxels to stride along the ``axis`` dimension between
+            slab projections.
+        projection_type: Type of intensity projection. Possible inputs are
+            ``'max'`` (the default), ``'min'``, ``'mean'``, ``'median'``, or
+            ``'quantile'``. If ``'quantile'`` is used, ``q`` must also be
+            supplied.
+        q: Quantile to use for intensity projections. This argument is required
+            if ``projection_type`` is ``'quantile'`` and is silently ignored
+            otherwise.
+        full_slabs_only: Boolean. Should projections be done only for slabs
+            that are ``slab_thickness`` thick? Default is ``True``.
+            If ``False``, some slabs may not be ``slab_thickness`` thick
+            depending on the size of the image, slab thickness, and stride.
+
+    Example:
+        >>> import torchio as tio
+        >>> sub = tio.datasets.Colin27()
+        >>> axial_mips = tio.Projection("S", slab_thickness=20)
+        >>> sub_t = axial_mips(sub)
+        >>> sub_t.t1.plot()
+    """
+    def __init__(
+            self,
+            axis: str,
+            slab_thickness: Optional[int] = None,
+            stride: Optional[int] = 1,
+            projection_type: Optional[str] = 'max',
+            q: Optional[float] = None,
+            full_slabs_only: Optional[bool] = True,
+            **kwargs
+            ):
+        super().__init__(**kwargs)
+        self.args_names = (
+            'axis', 'slab_thickness', 'stride',
+            'projection_type', 'q', 'full_slabs_only'
+            )
+        self.axis = axis
+        self.slab_thickness = slab_thickness
+        self.stride = stride
+        self.projection_type = projection_type
+        self.q = q
+        self.full_slabs_only = full_slabs_only
+        self.projection_fun = self.get_projection_function()
+
+    def get_projection_function(self):
+        if self.projection_type == 'max':
+            projection_fun = torch.max
+        elif self.projection_type == 'min':
+            projection_fun = torch.min
+        elif self.projection_type == 'mean':
+            projection_fun = torch.mean
+        elif self.projection_type == 'median':
+            projection_fun = torch.median
+        elif self.projection_type == 'quantile':
+            projection_fun = torch.quantile
+            self.validate_quantile()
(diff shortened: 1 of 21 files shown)


fepegar:
Thanks, @mattwarkentin. This looks nice. I've worked on vessel segmentation so I appreciate the importance of these techniques.

Before I look into anything, could you please post some images?

mattwarkentin:
Here is an example chest CT scan before (original) and after axial maximum intensity projection via `tio.Projection()`:
Screen Shot 2021-08-18 at 5 07 01 PM
Screen Shot 2021-08-18 at 5 07 11 PM

mattwarkentin:
Minimum intensity projection:
Screen Shot 2021-08-18 at 5 11 05 PM

mattwarkentin:
I have noticed a couple bugs while trying to produce some images. Will work on a fix.

mattwarkentin:
Pretty happy with the state of the PR at this point. I will leave it be now until you've had a chance to review!

torchio/transforms/preprocessing/intensity/projection.py
```diff
@@ -0,0 +1,158 @@
+import torch
+from torchio.data.image import ScalarImage
+from ....data.subject import Subject
+from ...intensity_transform import IntensityTransform
+from typing import Optional
+from math import ceil
+
+
+class Projection(IntensityTransform):
```
fepegar, line 9: What about `SlabProjection`? `Projection` doesn't feel very descriptive.

torchio/transforms/preprocessing/intensity/projection.py
```diff
@@ -0,0 +1,158 @@
...
+    """Project intensities along a given axis, possibly with sliding slabs.
+
+    Args:
+        axis: Possible inputs are ``'Left'``, ``'Right'``, ``'Anterior'``,
```
fepegar, line 13: For many images, anatomical axes don't make sense. It should be possible to pass the axis index instead (0, 1 or 2). I'll add a comment on this below*

torchio/transforms/preprocessing/intensity/projection.py
```diff
@@ -0,0 +1,158 @@
...
+        else:
+            raise ValueError(message)
+
+    def apply_transform(self, subject: Subject) -> Subject:
+        for image in self.get_images(subject):
+            self.apply_projection(image)
+        return subject
+
+    def apply_projection(self, image: ScalarImage) -> None:
+        self.axis_index = image.axis_name_to_index(self.axis)
```
fepegar, line 115: *Maybe `axis_name_to_index` could be modified to just return the input if it is 0, 1 or 2.

torchio/transforms/preprocessing/intensity/projection.py
```diff
@@ -0,0 +1,158 @@
+import torch
+from torchio.data.image import ScalarImage
+from ....data.subject import Subject
+from ...intensity_transform import IntensityTransform
+from typing import Optional
+from math import ceil
```
fepegar, line 6: Maybe we can use `torch.ceil` to avoid this import.

torchio/transforms/preprocessing/intensity/projection.py
```diff
@@ -0,0 +1,158 @@
...
+        axis: Possible inputs are ``'Left'``, ``'Right'``, ``'Anterior'``,
+                ``'Posterior'``, ``'Inferior'``, ``'Superior'``. Lower-case
+                versions and first letters are also valid, as only the first
+                letter will be used.
+        slab_thickness: Thickness of slab projections. In other words, the
+            number of voxels in the ``axis`` dimension to project across.
+            If ``None``, the projection will be done across the entire span of
+            the ``axis`` dimension (i.e. ``axis`` dimension will be reduced to
+            1).
+        stride: Number of voxels to stride along the ``axis`` dimension between
```
fepegar, line 22: When is this useful?

torchio/transforms/preprocessing/intensity/projection.py
```diff
@@ -0,0 +1,158 @@
...
+            slab projections.
+        projection_type: Type of intensity projection. Possible inputs are
+            ``'max'`` (the default), ``'min'``, ``'mean'``, ``'median'``, or
+            ``'quantile'``. If ``'quantile'`` is used, ``q`` must also be
+            supplied.
+        q: Quantile to use for intensity projections. This argument is required
```
fepegar, line 28: Can we switch to percentiles, for consistency with `RescaleIntensity`?

torchio/transforms/preprocessing/intensity/projection.py
```diff
@@ -0,0 +1,158 @@
...
+
+    .. plot::
+
+        import torchio as tio
+        sub = tio.datasets.Slicer('CTChest')
+        ct = sub.CT_chest
+        axial_mips = tio.Projection("S", slab_thickness=20, stride=20)
+        ct_mips = axial_mips(ct)
+        sub.add_image(ct_mips, 'MIP')
+        sub.plot()
```
fepegar, line 51: Nice! It's a shame the contrast is so low. There's a new transform that could make this look prettier...

torchio/transforms/preprocessing/intensity/projection.py
```diff
@@ -0,0 +1,158 @@
...
+
+    """
+    def __init__(
+            self,
+            axis: str,
+            slab_thickness: Optional[int] = None,
+            stride: Optional[int] = 1,
```
fepegar, line 58: `Optional` means the argument may be `None`. In this case, the typing hint should just be `int`. Same goes for the other kwargs.

torchio/transforms/preprocessing/intensity/projection.py
```diff
@@ -0,0 +1,158 @@
...
+        super().__init__(**kwargs)
+        self.args_names = (
+            'axis', 'slab_thickness', 'stride',
+            'projection_type', 'q', 'full_slabs_only'
+            )
+        self.axis = axis
+        self.slab_thickness = slab_thickness
+        self.stride = stride
+        self.projection_type = projection_type
+        self.q = q
```
fepegar, line 73: Can we use a more descriptive variable name? E.g. `quantile` (or `percentile`). By the way, if we do change the name of the argument to `percentile`, you can just divide by 100 here so the rest of the code stays the same.

torchio/transforms/preprocessing/intensity/projection.py
```diff
@@ -0,0 +1,158 @@
...
+            projection_fun = torch.amax
+        elif self.projection_type == 'min':
+            projection_fun = torch.amin
+        elif self.projection_type == 'mean':
+            projection_fun = torch.mean
+        elif self.projection_type == 'median':
+            projection_fun = torch.median
+        elif self.projection_type == 'quantile':
+            projection_fun = torch.quantile
+            self.validate_quantile()
```
fepegar, line 88: Would it make sense to move this to `__init__`? So errors are caught as early as possible.

torchio/transforms/preprocessing/intensity/projection.py
```diff
@@ -0,0 +1,158 @@
...
+            message = (
+                '`projection_type` must be one of "max", "min", "mean",'
+                ' "median", or "quantile".'
+                )
+            raise ValueError(message)
+        return projection_fun
+
+    def validate_quantile(self):
+        message = (
+            'For `projection_type="quantile"`, `q` must be a scalar value'
```
fepegar, line 99: Markdown formatting doesn't work for error messages :)

torchio/transforms/preprocessing/intensity/projection.py
fepegar: Note that we would need to add 1 to the axis index, because e.g. 0 would be the second dimension (as the channels one is actually the first one).

fepegar: Also, the `spatial_shape` property of the image might be useful.

torchio/transforms/preprocessing/intensity/projection.py
```diff
@@ -0,0 +1,158 @@
...
+                start_index += self.stride
+        else:
+            num_slabs = ceil(self.axis_span / self.stride)
+
+        slabs = []
+        start_index = 0
+        end_index = start_index + self.slab_thickness
+
+        for _ in range(num_slabs):
+            slab_indices = torch.tensor(list(range(start_index, end_index)))
```
fepegar, line 141: I think you can use `torch.arange` for this.

fepegar (changes requested):
Looking great! I've added some comments. Let me know if you have any questions.

mattwarkentin:
Thanks for the feedback. I am fairly new to `python` (I come from the `R` world, primarily) so I welcome the feedback to improve the PR and my own coding best-practices. I will respond to each of your in-line comments.

torchio/transforms/preprocessing/intensity/projection.py
mattwarkentin: I originally used `IntensityProjection` but thought it was redundant since it was an intensity transform. Happy to try something different.

torchio/transforms/preprocessing/intensity/projection.py
mattwarkentin: Do you want me to take a look at modifying `axis_name_to_index()`, or just implement a method in `Projection` to validate the chosen axis exists and add 1 to it for indexing purposes?

torchio/transforms/preprocessing/intensity/projection.py
mattwarkentin: Which part are you referring to? The projection across the whole dimension or the stride argument? I assume the former. I think the projection across the entire axis is the default behaviour of something like SimpleITK's `sitk.MaximumProjection()` method. I would have to think about it some more, but I think projection across the whole dimension is probably only useful if you've already cropped down the image, otherwise you are projecting across such a large distance that the resulting image probably doesn't make much sense structurally. I'm definitely open to changing this default.

torchio/transforms/preprocessing/intensity/projection.py
mattwarkentin: Ah okay, good to know. I am new to python so I am learning many of these conventions on the fly.

torchio/transforms/preprocessing/intensity/projection.py
mattwarkentin: `self.get_projection_function` gets called in `__init__` so `self.validate_quantile` does get called down the stack, but only if `projection_type='quantile'`. I did this so that `q` could be ignored but I could also just change `validate_quantile` to `pass` if `projection_type` is not `'quantile'` and it would be more obvious that it would throw errors during init.

torchio/transforms/preprocessing/intensity/projection.py
mattwarkentin: I actually do know this! This type of error message formatting is a carry-over from how we often present error messages in `R`. Will change to be more pythonic.

torchio/transforms/preprocessing/intensity/projection.py
mattwarkentin: After re-reading, I think maybe you are asking about `stride` parameter. Since projections are summarizing transforms of the original image, if `stride < slab_thickness`, for example, the slices in the new image contain structurally overlapping information from the original image.

Setting `stride = slab_thickness` will ensure the new image image is a summary of the old image without overlapping. Does that make sense? Maybe stride should just always be equal to slab thickness and it shouldn't be a configurable argument, unless we can think of good reasons to support strides < or > slab thickness.

mattwarkentin: FWIW, in [this](https://youtu.be/Qmdl0zckFnw?t=367) they talk about moving their slab less than the slab thickness, but perhaps this is only valuable for visualization tools.

torchio/transforms/preprocessing/intensity/projection.py
fepegar: I'll think of the best approach and maybe commit something.

torchio/transforms/preprocessing/intensity/projection.py
fepegar: Yes, I was referring to the stride. Unfortunately, GitHub renders 4 lines when you click on only one (the last one).

I guess my question is "when would a stride > 1 be useful?". It's nice that the user has the choice, but I wonder if it adds unnecessary complexity to the transform. Do you use this argument in your application?

torchio/transforms/preprocessing/intensity/projection.py
fepegar: Makes sense. But I think the validation is cheap enough that it can be computed at instantiation time, to avoid frustrating runtime errors (e.g. once training has started).

torchio/transforms/preprocessing/intensity/projection.py
mattwarkentin: Speed considerations might be one reason. If you have a fairly high-resolution image, like a CT scan with dimensions `(512, 512, 300)` and spacing of `(0.7, 0.7, 1.0)` and want 10mm slab projections (`slab_thickness=10`), then striding only 1 in the axial dimension is probably overkill as it only moves 1mm in physical space. Striding 2 (or more) would halve the number of slab computations with (probably) no meaningful loss in structure representation (of course this is context dependent, my brain goes to chest CTs).

fepegar: Ok, that makes sense. Would it be worth it to modify the affine accordingly? I'd need to do the maths, but I guess the spacing along the specified axis could be modified so the image stays in a similar position in world coordinates.

mattwarkentin: Hmm, yes, I suppose it would make sense to update the affine. Should be straightforward enough, I think. Modify the affine value for the chosen `axis` to be `value * self.slab_thickness`?? Something like that.

mattwarkentin:
One thing I have been thinking about and would like your feedback on...currently slabs are handled differently at the start versus the end of the axis dimension. I don't like this behaviour.

Proposal:

- If `full_slabs_only=True`, we figure out how many slabs can possibly be fit in the image and "center" the set of slabs in the image (i.e. equal spacing/padding on the start and end dimension to fit max possible slabs)
- If `full_slabs_only=False`, we start projecting as far outside the starting side dimension as we will end up outside the ending side of the dimension. This way the "slabbing" is always handled equivalently on both "ends" of the axis.

Thoughts?

torchio/transforms/preprocessing/intensity/projection.py
fepegar: I might be wrong, but I think multiplying by the stride (not thickness) would be the solution.

fepegar:
I didn't understand your last comment very well. I guess it's an implementation detail and trust you to handle it as you consider convenient.

torchio/transforms/preprocessing/intensity/projection.py
fepegar: I think the spacing should change with the stride, not with the slab thickness. The current implementation should touch the origin as well, as shown with the results from your snippet:

SlicerCapture

As you see, the image seems to "move forward" whereas it should approximately stay in the same physical space.

I might have screwed up the axes with my changes, because the projection seems to have happened along the A-P axis, instead of the I-S.

fepegar: > I might have screwed up the axes with my changes, because the projection seems to have happened along the A-P axis, instead of the I-S.

This is because `slab = tensor.index_select(axis_index, slab_indices)` is not taking into account the channels dimension, I'll commit a fix.

romainVala: about the affine change, 
the spacing is the distance (in mm) between two slices, so I would say
original_slice_spacing * stride
note also that the translation part should also be adapted, given the fact there is an implicit croping (compare to the original images) if full_slabs_only is True (same way as affine is modify wen croping)

fepegar:
I think the PR wasn't affected. I'm sorry the branch renaming messed with your repo! Maybe it'd help to follow [these instructions](https://github.com/fepegar/torchio/issues/716#issuecomment-953919047) or start with a fresh clone, I'm not sure. This was almost ready, I can maybe give it a try one of these evenings.

mattwarkentin:
I think it was my fault for deleting and re-forking, rather than just renaming `master -> main`. The PR itself was probably fine, I just acted too quickly without thinking about how I might break my connection to this PR.

PR #627, opened 2021-08-20
Fix default pad value for label maps

### LINKED ISSUE #626: Pad operation modifies labels
**🐛Bug**
The Affine transformation modifies labels at the edges.

**To reproduce**

```python
import torchio as tio
import torch
a = torch.tensor([[1, 0, 0],[0, 1, 0],[0, 0, 1]]).reshape(1, 3, 3, 1)
s = tio.Subject(lab=tio.LabelMap(tensor=a))
cp = tio.RandomAffine(translation = (0, 1, 1), default_pad_value='otsu')
sb = cp(s)
print(sb.lab.data)

```

**Expected behavior**

The augmented map is now corrupted at the edges after the translation.
The default padding value should always be 0 for labels.

**Actual behavior**

At the moment the values from outside of the original data range are set to some non-zero value.
This is expected for the non-label maps, but is a corruption of data for the labels.

fepegar:
Fixes #626.

**Description**
If the Otsu technique is used to compute the default padding value for the affine augmentation, it should only be used for scalar images, as noted by @glupol in #626.

**Checklist**

### DIFF
--- a/tests/transforms/augmentation/test_random_affine.py
+++ b/tests/transforms/augmentation/test_random_affine.py
@@ -1,4 +1,5 @@
-from torchio.transforms import RandomAffine
+import torch
+import torchio as tio
 from ...utils import TorchioTestCase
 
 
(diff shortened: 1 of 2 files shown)


fepegar:
@all-contributors please add @glupol for bug

PR #640, opened 2021-08-30
Add support to pass shape and affine to Resample

### LINKED ISSUE #638: Resample transform should not need Image attribut
**🐛Bug**

This is not really a bug, but a memory issue, of the transform object that can store one or two `torchio.Image` object
This is dependant on the input argument choosen

If target is the string of a subject image's name, then it is fine.
Other case (path to a file) or tio.Image, then the object store internally an Image (or sometimes 2, one in `target` and one in `reference_image` attribute)

**To reproduce**

```python
import torchio as tio
import pickle
s=tio.datasets.Colin27()
t=tio.Resample(target='t1')
st = t(s)

#save the object os pickle to see the real size
pickle.dump(t,open('/tmp/test.p','wb'))
!du -sh /tmp/test.p
# 4.0K	/tmp/test.p

t=tio.Resample(target='/tmp/t1.nii')
st = t(s)
pickle.dump(t,open('/tmp/test.p','wb'))
!du -sh /tmp/test.p
# 28M	/tmp/test.p

t.target
Out[32]: '/home/romain.valabregue/t1.nii'

t.reference_image
Out[33]: ScalarImage(shape: (1, 182, 218, 182); spacing: (1.00, 1.00, 1.00); orientation: LAS+; memory: 27.5 MiB; dtype: torch.FloatTensor)


```

**Expected behavior**
In order to perform a Resample, there is no need of the reference image voxel. 
(There is a common misunderstanding between coregistration and resample)
Coregistration will need the reference image voxel to be load, (as the similarity metric need to be compute)
But Resample only need the reference's bonding box and reference's affine

nibabel implementation of [resample](https://nipy.org/nibabel/reference/nibabel.processing.html#nibabel.processing.resample_from_to
), take a list  (shape, affine) (or an object that have the both attribute)

**Actual behavior**
I am not very used with sitk, but since this software is intended to to  coregistration may be this is  why you need to construct a sitk image, in order to apply the resample

I get into trouble, because my application is keeping a list of transform's history during a training epoch. (So if an epoch contains 1000 iteration and each iteration store a resample, I end up storing 1000 3D matrix of the same reference image ...

and indeed I get memory issue (increasing cpu memory during the iteration) 

Having `tio.Image` as argument could be find if it is only used to get shape and affine, (and so that the data is never loaded)

It is not totally clear to me, when exactly the Image voxel is being load into memory.  

Actually I was using the `RandomAnisotropy` transform, that create 2 `Resample` transform the second one taking as `target` argument directly an `Image`.
It seems to me that this case is forcing the voxel data to be load into memory ...

Since I am 100% sure I did obverse a constant memory increase during my training : The RandomAnisotropy call do induce memory load of reference_image and or target attributes
which should be avoid, for Resampling purpose

fepegar:
Fixes #638.

**Checklist**

### DIFF
--- a/torchio/transforms/preprocessing/spatial/resample.py
+++ b/torchio/transforms/preprocessing/spatial/resample.py
@@ -6,18 +6,14 @@ import torch
 import numpy as np
 import SimpleITK as sitk
 
-from ....data.io import sitk_to_nib
+from ....data.io import sitk_to_nib, get_sitk_metadata_from_ras_affine
 from ....data.subject import Subject
-from ....typing import TypeTripletFloat
+from ....typing import TypeTripletFloat, TypePath
 from ....data.image import Image, ScalarImage
 from ... import SpatialTransform
 
 
 TypeSpacing = Union[float, Tuple[float, float, float]]
-TypeTarget = Tuple[
-    Optional[Union[Image, str]],
-    Optional[Tuple[float, float, float]],
-]
 
 
 class Resample(SpatialTransform):
@@ -36,7 +32,7 @@ class Resample(SpatialTransform):
             transform. See usage in the example below.
         image_interpolation: See :ref:`Interpolation`.
         scalars_only: Apply only to instances of :class:`~torchio.ScalarImage`.
-            See :class:`~torchio.transforms.RandomAnisotropy`.
+            Used internally by :class:`~torchio.transforms.RandomAnisotropy`.
         **kwargs: See :class:`~torchio.transforms.Transform` for additional
             keyword arguments.
 
@@ -74,11 +70,12 @@ class Resample(SpatialTransform):
             ):
         super().__init__(**kwargs)
         self.target = target
-        self.reference_image, self.target_spacing = self.parse_target(target)
-        parsed_interpolation = self.parse_interpolation(image_interpolation)
-        self.image_interpolation = parsed_interpolation
+        self.image_interpolation = self.parse_interpolation(
+            image_interpolation)
         self.pre_affine_name = pre_affine_name
         self.scalars_only = scalars_only
+        self.target_shape = None
+        self.target_affine = None
         self.args_names = (
             'target',
             'image_interpolation',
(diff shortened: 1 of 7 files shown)


fepegar:
TODO: update docstring.

@romainVala Can you try your snippet in #638 and see if this is better?

romainVala:
just get a quick look !
amazing quite a lot of changes and the result seems to be ok for me ! 
the pickel is ok, and the history looks as expected

great job, i'll look more to better understand, but difficult to review

torchio/transforms/preprocessing/spatial/resample.py
```diff
@@ -74,41 +70,21 @@ def __init__(
...
         self.target = target
-        self.reference_image, self.target_spacing = self.parse_target(target)
-        parsed_interpolation = self.parse_interpolation(image_interpolation)
-        self.image_interpolation = parsed_interpolation
+        self.image_interpolation = self.parse_interpolation(
+            image_interpolation)
         self.pre_affine_name = pre_affine_name
         self.scalars_only = scalars_only
+        self.target_shape = None
+        self.target_affine = None
```
romainVala, line 78: not used ?

torchio/transforms/preprocessing/spatial/resample.py
```diff
@@ -183,64 +161,122 @@ def apply_transform(self, subject: Subject) -> Subject:
...
+        # 5) A tuple of shape, affine
+        # The fourth case is the different one
+        if isinstance(target, (str, Path, Image)):
+            if Path(target).is_file():
+                # It's an existing file
+                path = target
+                image = ScalarImage(path)
+            elif isinstance(target, Image):
+                # It's a TorchIO image
+                image = target
```
romainVala, line 203: it was used before by RandomAnisotropy, but I wonder if still usefull ?

romainVala (commented):
looks good

torchio/transforms/preprocessing/spatial/resample.py
fepegar: You're right, good catch! I was trying a different design and this is a trace of that.

torchio/transforms/preprocessing/spatial/resample.py
fepegar: Hmm I think it's still good to leave it, especially for backward compatibility.

romainVala: sure no problem, I was just wondering if you see other usecase

fepegar:
Thanks for reporting and reviewing, @romainVala!

PR #682, opened 2021-10-05
Add plot_volume indices parameter

laynr:
Fixes https://github.com/fepegar/torchio/discussions/681

**Description**
Two changes:
- Add plot_volume indices parameter that defaults to None
- Close figure window with `plt.close(figure_object)`  as described in [documentation](https://matplotlib.org/3.1.1/api/_as_gen/matplotlib.pyplot.close.html) and described  [here](https://stackoverflow.com/questions/9622163/save-plot-to-image-file-instead-of-displaying-it-using-matplotlib).  Doing this because even when `show=True` the image would show and I don't have  lots of open figures during a loop.

**Checklist**

note done.

### DIFF
--- a/torchio/visualization.py
+++ b/torchio/visualization.py
@@ -155,6 +155,7 @@ def plot_subject(
         fig.savefig(output_path)
     if show:
         plt.show()
+    plt.close(fig) 
 
 
 def color_labels(arrays, cmap_dict):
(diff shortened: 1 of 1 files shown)


fepegar:
Hi, @laynr. Why did you close this?

PR #718, opened 2021-10-26
Add support to copy an object that inherits from `Subject`

### LINKED ISSUE #694: Use of self.__class__ in __copy__ method of Subject
**🚀 Feature**
possibility to inherit torchio Subject

**Motivation**

In our current workflow we want to build on top of the torchio framework. We have subjects with 2 mandatory properties, a study name and a patient id. Therefore we have a subject class that looks like this:

```python
class Subject(tio.Subject):

    def __init__(self, study_name: str, patient_id: str, *args, **kwargs):
        kwargs['study_name'] = study_name
        kwargs['patient_id'] = patient_id
        super().__init__(*args, **kwargs)
```
**Pitch**

What I would like to see is that this method:

```python
    def __copy__(self):
        result_dict = {}
            if isinstance(value, Image):
                value = copy.copy(value)
            else:
                value = copy.deepcopy(value)
            result_dict[key] = value
        new = Subject(result_dict)
        new.applied_transforms = self.applied_transforms[:]
        return new
```
is replaced by

```python
    def __copy__(self):
        result_dict = {}
            if isinstance(value, Image):
                value = copy.copy(value)
            else:
                value = copy.deepcopy(value)
            result_dict[key] = value
        new = self.__class__(**result_dict)
        new.applied_transforms = self.applied_transforms[:]
        return new
```
to make custom subject classes that inherit torchio Subject compatible with the other components in the framework. The Image class has already a  __copy__ method that uses __class__ property.

**Alternatives**

The alternative is overwriting the __copy__ method in the custom classes that inherit torchio subject. However, this seems less future proof.

dmus:
Fixes #694.

**Description**
use of __class__ in __copy__ method of subject to make subclass Subject possible

**Checklist**

### DIFF
--- a/tests/data/test_subject.py
+++ b/tests/data/test_subject.py
@@ -106,3 +106,14 @@ class TestSubject(TorchioTestCase):
     def test_no_images(self):
         with self.assertRaises(ValueError):
             tio.Subject(a=0)
+
+    def test_copy_subclass(self):
+        class CustomSubject(tio.Subject):
+            def __init__(self, img: tio.Image, *args, **kwargs):
+                kwargs['img'] = img
+                super().__init__(*args, **kwargs)
+
+        subject = CustomSubject(img=tio.ScalarImage(tensor=torch.rand(1, 3, 3, 4)))
+        new_subject = copy.copy(subject)
+
+        assert isinstance(new_subject, CustomSubject)
--- a/tests/data/test_subject.py
+++ b/tests/data/test_subject.py
@@ -112,8 +112,18 @@ class TestSubject(TorchioTestCase):
             def __init__(self, img: tio.Image, *args, **kwargs):
                 kwargs['img'] = img
                 super().__init__(*args, **kwargs)
-
-        subject = CustomSubject(img=tio.ScalarImage(tensor=torch.rand(1, 3, 3, 4)))
+        tensor = torch.rand(1, 3, 3, 4)
+        subject = CustomSubject(img=tio.ScalarImage(tensor=tensor))
         new_subject = copy.copy(subject)
+        assert isinstance(new_subject, CustomSubject)
 
+    def test_copy_subclass_no_kwargs(self):
+        class CustomSubject(tio.Subject):
+            def __init__(self):
+                tensor = torch.rand(1, 3, 3, 4)
+                img = tio.ScalarImage(tensor=tensor)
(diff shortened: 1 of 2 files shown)


torchio/data/subject.py
```diff
@@ -71,7 +71,7 @@ def __copy__(self):
             else:
                 value = copy.deepcopy(value)
             result_dict[key] = value
-        new = Subject(result_dict)
+        new = self.__class__(**result_dict)
```
justusschock, line 74: Why can't you simply use the dict as before here?

tests/data/test_subject.py
```diff
@@ -106,3 +106,24 @@ def test_bad_arg(self):
...
+    def test_copy_subclass_no_kwargs(self):
+        class CustomSubject(tio.Subject):
+            def __init__(self):
+                tensor = torch.rand(1, 3, 3, 4)
+                img = tio.ScalarImage(tensor=tensor)
+                kwargs = {'img': img}
+                super().__init__(**kwargs)
+        subject = CustomSubject()
+        new_subject = copy.copy(subject)
+        assert isinstance(new_subject, CustomSubject)
```
justusschock, line 129:
```suggestion
    def test_copy_subclass(self):
        class CustomSubject(tio.Subject):
            pass
        assert isinstance(new_subject, CustomSubject)
```

IMO overything else (including argument forwarding) tests the correct implementation of the subclass which is up to the user

justusschock (approved):

justusschock:
+1 on this.

@fepegar what do you think on my arguments on the tests above? For me they succeeded locally.

torchio/data/subject.py
justusschock, line 74:
```suggestion
        new = type(self)(**result_dict)
```
is also a bit cleaner :)

fepegar:
I'm not sure this can be solved easily. @justusschock, if you have a branch with your proposed changes, we can test that one. I don't think we should remove the current tests, though.

fepegar:
Maybe it would be easier to override the `__copy__` method, but I'm not sure.

justusschock:
@fepegar I opened a similar PR with more minimal tests in #794

Overriding the `__copy__` could definitely work, but it would be a lot of copy-paste because you'd need to add all the other logic as well :)

fepegar:
Closed in favor of #794.

PR #727, opened 2021-10-31
Add method to show image with external viewer

### LINKED ISSUE #726: Very lightweight interactive viewer
**🚀 Feature**

Hi @fepegar,

I love the new `Image.to_gif()` feature. It makes it really easy to create some cool visualizations. However, I would still love there to be some (very) lightweight interactive viewer. It is definitely a rabbit hole to implement anything too complex, and there are already good external viewer solutions, but I would love to be able to have a simple slider to flip through slices (along a specific axis). Maybe we can consider a new method `view()`.

```py
img = tio.ScalarImage(...)
img.view(axis=2) # one of 0, 1, or 2; possibly anatomical labels like 'axial', 'sagittal', 'coronal'
```

I don't currently know enough about the best way to implement such a slider plotting interface, but if there is interest I would be happy to look into this further and perhaps contribute a PR. What are your thoughts?

**Motivation**

It would make me happy?? :)

**Alternatives**

A very low-burden solution might be to use the `Show()` function from `SimpleITK`. This will open the image in an external viewer of choice (e.g. Fiji, ITKSnap, 3D Slicer, ImageJ, etc.). This would be fairly easy to implement with existing infrastructure. In fact, this might be useful to add independent of possibly implementing an in-house interactive viewer. Thoughts?

```py
class Image(...):
    ...
    def show(self, title, debugOn):
        sitk.Show(self.as_sitk(), title, debugOn)
```

fepegar:
Resolves #726.

@mattwarkentin can you please review this?

**Description**
This PR adds a method in `Image` to open the image using an external viewer such as ITK-SNAP or 3D Slicer.

```python
import torchio as tio
t1 = tio.datasets.Sheep().t1
t1.show()
```

Other viewers that could be tried automatically:
- [FSLeyes](https://fsl.fmrib.ox.ac.uk/fsl/fslwiki/FSLeyes)
- [mrview](https://mrtrix.readthedocs.io/en/latest/reference/commands/mrview.html)
- [ImageJ](https://imagej.net/software/imagej2/) / [Fiji](https://imagej.net/software/fiji/) (these might be interesting for 2D or histology stacks)

I don't have fast access to Windows, so I'm not sure what the paths would look like.

I don't think the paths on Linux could be easily guessed. I guess we could try just the name of the binary, which would probably be in the `PATH`.

**Checklist**

### DIFF
--- a/torchio/data/image.py
+++ b/torchio/data/image.py
@@ -665,7 +665,6 @@ class Image(dict):
         image_viewer = sitk.ImageViewer()
         # This is so that 3D Slicer creates segmentation nodes from label maps
         if self.__class__.__name__ == 'LabelMap':
-            print('Im here')
             image_viewer.SetFileExtension('.seg.nrrd')
         if viewer_path is not None:
             image_viewer.SetApplication(str(viewer_path))
--- a/torchio/data/image.py
+++ b/torchio/data/image.py
@@ -647,7 +647,7 @@ class Image(dict):
             from ..visualization import plot_volume  # avoid circular import
             plot_volume(self, **kwargs)
 
-    def show(self, viewer_path=None):
+    def show(self, viewer_path: Optional[TypePath] = None) -> None:
         """Open the image using external software.
 
         Args:
--- a/torchio/data/image.py
+++ b/torchio/data/image.py
@@ -10,7 +10,7 @@ import nibabel as nib
 import SimpleITK as sitk
 from deprecated import deprecated
 
-from ..utils import get_stem
+from ..utils import get_stem, guess_external_viewer
 from ..typing import (
     TypeData,
     TypePath,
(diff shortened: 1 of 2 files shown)


mattwarkentin (approved):
LGTM!

mattwarkentin:
Actually, I might have a suggestion, one sec...

torchio/data/image.py
```diff
@@ -647,6 +647,38 @@ def plot(self, **kwargs) -> None:
...
+                `ITK-SNAP <http://www.itksnap.org/pmwiki/pmwiki.php>`_ and
+                `3D Slicer <https://www.slicer.org/>`_.
+
+        Raises:
+            RuntimeError: If the viewer is not found.
+        """
+        sitk_image = self.as_sitk()
+        image_viewer = sitk.ImageViewer()
+        if viewer_path is not None:
+            image_viewer.SetApplication(str(viewer_path))
```
mattwarkentin, line 667: Following from [this](https://discourse.slicer.org/t/loading-image-and-corresponding-mask-on-slicer-from-command-line/9930/2) thread, maybe we could consider using something like:

```py
if self.type == 'label':
    image_viewer.SetFileExtension('.label.nii')
```

This will ensure that the `LabelMap` will actually get loaded as a label map in Slicer, rather than as a normal image volume and should have no side-effect for other software. Thoughts?

mattwarkentin (changes requested):

torchio/data/image.py
mattwarkentin: Slicer looks for the `.label` or `.seg` to the extension to try and identify label maps/segmentations when automatically loading from the command-line.

fepegar: Good idea! It doesn't seem to work with `label`, but I tried `seg.nrrd` and it's ok.

torchio/data/image.py
```diff
@@ -647,6 +647,42 @@ def plot(self, **kwargs) -> None:
...
+                `3D Slicer <https://www.slicer.org/>`_.
+
+        Raises:
+            RuntimeError: If the viewer is not found.
+        """
+        sitk_image = self.as_sitk()
+        image_viewer = sitk.ImageViewer()
+        # This is so that 3D Slicer creates segmentation nodes from label maps
+        if self.__class__.__name__ == 'LabelMap':
+            print('Im here')
```
mattwarkentin, line 668: `print('Not for long!')`

mattwarkentin (changes requested):

torchio/data/image.py
fepegar: Hahahah `pre-commit` should have detected that and not allowed me to commit. There's obviously something wrong so this is a good catch. Thanks!

fepegar:
I have now added and tested some code to guess paths to viewers on Windows and Linux. It's a bit messy because I know nothing about OSs, but it should be convenient enough for most users, which I guess is in the spirit of TorchIO.

PR #763, opened 2021-11-29
Fix affine getter for multipath images

### LINKED ISSUE #762: Error when using a list of paths in Image class

### Bug summary

See reproducible example, it gives an RuntimeError: Exception thrown in SimpleITK ImageFileReader_ReadImageInformation

### Code for reproduction

```python
path = '/tmp/tmpy168nb8s/Results/test_cvf/20211129_180113_860_3QC1RFRH/saved_images/prediction_test_cvf_001.mhd'
prediction = tio.Image(path=[path])

print(prediction.affine)
```

### Actual outcome

I would expect that this works without problems and gives a similar outcome when passing a single path instead of a list as the `path` argument

### Error messages

```python-traceback
File "reproduce.py", line 27, in <module>
    print(prediction.affine)
  File "/home/derk/rndeep/venv/lib/python3.6/site-packages/torchio/data/image.py", line 237, in affine
    affine = read_affine(self.path)
  File "/home/derk/rndeep/venv/lib/python3.6/site-packages/torchio/data/io.py", line 106, in read_affine
    reader = get_reader(path)
  File "/home/derk/rndeep/venv/lib/python3.6/site-packages/torchio/data/io.py", line 115, in get_reader
    reader.ReadImageInformation()
  File "/home/derk/rndeep/venv/lib/python3.6/site-packages/SimpleITK/SimpleITK.py", line 8673, in ReadImageInformation
    return _SimpleITK.ImageFileReader_ReadImageInformation(self)
```

### Expected outcome

I would expect it print the affine

### System info

```shell
Platform:   Linux-5.4.0-62-generic-x86_64-with-Ubuntu-18.04-bionic
TorchIO:    0.18.69
PyTorch:    1.10.0+cu102
SimpleITK:  1.2.4 (ITK 4.13)
NumPy:      1.19.5
Python:     3.6.9 (default, Jan 26 2021, 15:33:00) 
[GCC 8.4.0]
```

fepegar:
Fixes #762.

**Description**
An error was being raised when trying to get the affine of multipath images, as the list of paths was being passed to the SimpleITK reader. Now, the images are loaded before returning the affine if the image is multipath.

**Checklist**

@dmus

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,3 +1,7 @@
+coverage:
+  status:
+    project: false
+
 ignore:
   - "setup.py"
   - "torchio/external/due.py"
--- a/tests/data/test_image.py
+++ b/tests/data/test_image.py
@@ -97,13 +97,16 @@ class TestImage(TorchioTestCase):
         with self.assertRaises(FileNotFoundError):
             tio.ScalarImage(path=['nopath', 'error'])
 
-    def test_with_a_list_of_paths(self):
+    def test_with_sequences_of_paths(self):
         shape = (5, 5, 5)
         path1 = self.get_image_path('path1', shape=shape)
         path2 = self.get_image_path('path2', shape=shape)
-        image = tio.ScalarImage(path=[path1, path2])
-        self.assertEqual(image.shape, (2, 5, 5, 5))
-        self.assertEqual(image[tio.STEM], ['path1', 'path2'])
+        paths_tuple = path1, path2
+        paths_list = list(paths_tuple)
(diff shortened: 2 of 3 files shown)


dmus:
Quick fix :) Now you load the image when it is a multipath image. Another option would be to only call read_affine with the first element of the list when it is a multipath? Or is that option not preferable?

fepegar:
Yes, that's another option. But I thought this would be safer, as it checks that all affines in the input images are the same (I think). I thought that this compromise of speed vs robustness was worth it as passing a sequence of paths is probably a not very common use case. Does that make sense?

PR #794, opened 2022-01-20
Fix copying of subclasses of Subject

### LINKED ISSUE #694: Use of self.__class__ in __copy__ method of Subject

justusschock:
Fixes #694 .

**Description**
Adds support (and tests) for copying subclasses.

Tests seem to be cleaner than #718 . Approach is almost the same!

**Checklist**


justusschock:
@fepegar can you re-run the tests ? It seems they failed to to connection errors...

fepegar:
> @fepegar can you re-run the tests ? It seems they failed due to connection errors...

Done, I've re-started the jobs.

---

@dmus, can you please review this PR?

fepegar:
@justusschock looking at [the build logs on RTD](https://readthedocs.org/projects/torchio/builds/15827887/), I think this implementation isn't working.

justusschock:
@fepegar I see that.

However,  After investigating a bit, I think this is because the pre-defined subjects (e.g. in `dataset`) break with the usual subject API by not accepting the images and other metadata as init arguments but only a year. This violates the liskov substitution principle. Usually the stuff happening there (like building the subject from a given version) would be a classmethod or a helper method but not part of the `__init__`.

fepegar:
Ok, thanks. I hadn't heard of the Liskov substitution principle. I'm not sure why the tests failed on ReadTheDocs but not in the GitHub action. Let's see if @dmus is happy with the changes and then we can merge.

fepegar:
> I'm not sure why the tests failed on ReadTheDocs but not in the GitHub action

This happens because RTD uses `Colin27` etc to create the gallery, but those are not called in the unit tests, which is problematic here. Maybe we should add a unit test that instantiates e.g. `Colin27`, to make sure this PR is compatible with those subclasses.

justusschock:
@fepegar to resolve this with colin27 etc. we could do one of two things: Resolve this by moving the dict building to a class method or introduce a special copy function there :) What would you prefer?

fepegar:
I'm not sure what each option would imply, but I'd rather not have to modify all the current subclasses with new methods.

justusschock:
@fepegar The first method would be the cleaner way in terms of API, but would however require changes from the user-side. 
There we would move all the code from the init-function to a classmethod like `from_year` which users would have to call then.

The second approach would be to copy-paste the current `__copy__` function to all the required subclasses (or at least a subset of them since for all the MNI things, only one copy would be necessary). This would then not require any additional user-interaction.

After looking at the code again, a third approach came to my mind:

Instead of having derived classes like `Colin27`, we could have a function with the same name that would just prepare the images dict and return it as an ordinary subject.

dmus:
changes look good. Other maybe related (or maybe another pr) question: when passing an Image or numpy array to a transform a temporary torchio Subject is created. Should it be possible to also have the custom Subject class there?

justusschock:
@dmus yes, I think this should just use the copy mechanism and that would be changed here. Will update after resolving the issues with the builtin datasets.

fepegar:
**Option 1: "moving the dict building to a class method"**
Would this break the current API? E.g. `colin = tio.datasets.Colin27()`.

**Option 2: "introduce a special copy function there"**
I feel like we shouldn't paste a lot of copies of a `__copy__` method in all the current subclasses.

**Option 3: "have a function with the same name that would just prepare the images dict and return it as an ordinary subject"**

Does this PR solve the original issue? Also, I think some current subclasses have some methods that would be lost. Maybe I have misunderstood this approach.

justusschock:
> Option 1: "moving the dict building to a class method"
Would this break the current API? E.g. colin = tio.datasets.Colin27().

Unfortunately yes. So while this would be the cleanest option from the programming side, I think breaking this API is off the table.

> Option 2: "introduce a special copy function there"
I feel like we shouldn't paste a lot of copies of a __copy__ method in all the current subclasses.

Well we could separate this into a helper function we just call if this is your main concern.


> Option 3: "have a function with the same name that would just prepare the images dict and return it as an ordinary subject"
Does this PR solve the original issue? Also, I think some current subclasses have some methods that would be lost. Maybe I have misunderstood this approach.

Yes it does, since this issue arises from the `__init__` of these subjects being different, which it wouldn't be any more.
We could also have subclasses that have everything and just a function calling those.

dmus:
How is progress going here?

dmus:
@justusschock @fepegar any updates here?

justusschock:
@dmus not from my side. Waiting on @fepegar to decide. So far I implemented the "proper" copying in a custom subclass which I don't really like, but is the current way to go

fepegar:
Apologies, guys. TorchIO has moved a bit in my list of priorities.
What is your preferred option?

justusschock:
Personally, I'd probably go with Option 2. Thoughts @dmus ?

dmus:
I do not have a strong preference, the most important is that the original issue will be solved.
I agree option 1 is the cleanest option, followed by option 3. Option 2 seems less clean, but keeps the API untouched if I understand it correctly. So if that is a hard requirement, that is the way to go

fepegar:
Option 2 it is, then.

> Well we could separate this into a helper function we just call if this is your main concern.

@justusschock, can you add that to this PR?

fepegar (approved):

fepegar:
I've built the docs, which make extensive use of these subclasses. Lacking better tests, I think this is good to go! Thanks a lot for your contributions, @justusschock and @dmus!

Issue #789, opened 2022-01-13
Samplers and Queues supporting different number of sampled patches per each subject of the SubjectsDataset


snipdome:
**🚀 Feature**
Samplers and Queues supporting different number of sampled patches per each subject of the SubjectsDataset

**Motivation**
In real case scenarios, the number of voxels of subjects may differ from each other. This may lead to torchio-users to want a different number of patches to be extracted from every subject volume. Right now, is only possible to have a fixed number of sampled patches for each subject.

**Additional context**
For CT and micro-CT volumes, subjects may have 100-10000 voxels in one of the three dimensions. Networks trained on such size need a patching system that should be able to support a great variety of volume sizes

fepegar:
Hi, @snipdome. I think this is a duplicate of
- #604
- #609 

Which I still need to address. We can follow the discussion in those. I'm going to close this for now. Feel free to reopen if I'm wrong.

PR #795, opened 2022-01-20
Add support to use different numbers of samples in the queue

fepegar:
Related to:
- #604 
- #609
- #789

**Description**
After these changes, one can decide how many patches to extract from each subject, using a `num_samples` kwarg when instantiating the subjects.

**Checklist**


jmlipman:
Hi Fernando.

I don't know whether I can download torchio and apply your pull request to try it locally, so, I could only read the code.

I like the idea of having a specific property to store the samples_per_volume (I was doing this, in practice). I have a question/concern. Consider the following scenario:

- Queue's size is 50.
- We are training on 2D slices of four 3D images.
- Each image has 30 slices. (total slices 30*4=120).

I have the feeling that the queue will be filled like this:

First fill: 30 slices of image 1 + 20 slices of image 2.
Second fill: 30 slices of image 3 + 20 slices of image 4.
Consequently, ignoring 10 slices from image 2 and 10 from image 4. This is because every time "fill" is called it loads a new subject without checking if we are done taking the all the slices from the last subject the previous time "fill" was called. At least I don't see how this is done, but, as I've mentioned before, I didn't run the code and I might be wrong.

fepegar:
> I don't know whether I can download torchio and apply your pull request to try it locally, so, I could only read the code.

Maybe you can use

```shell
git clone -b 604-queue-different-samples git@github.com:fepegar/torchio.git
```

About the other concerns, I think that using a much larger queue length would make the stochastic process less sensitive to this issue. We might get fewer patches from a specific subject during an epoch, but that shouldn't happen in the next one if the subjects list is shuffled (which it is, by default). If you still think this is a concern, we can look into it and find alternatives.

PR #841, opened 2022-03-14
Fix the wrong mapping in sequential labels transform

### LINKED ISSUE #840: SequentialLabels generates a wrong mapping

### Bug summary

> For example, if a label map has 6 labels with IDs (3, 5, 9, 15, 16, 23), then this will apply a RemapLabels transform with remapping={3: 1, 5: 2, 9: 3, 15: 4, 16: 5, 23: 6}. 

The documentation describes that the behavior of SequentialLabels is to map (3, 5, 9, 15, 16, 23) to (1, 2, 3, 4, 5, 6). But when I use SequentialLabels, I get (3, 1, 2, 3, 4, 5)

And I have read the source code, `SequentialLabels` uses the code below to generate mapping.

```python
unique_labels = torch.unique(image.data)
remapping = {
     unique_labels[i].item(): i
     for i in range(1, len(unique_labels))
}
```

When we input the (3, 5, 9, 15, 16, 23), the mapping is {5: 1, 9: 2, 15: 3, 16: 4, 23: 5}, so the 3 is reused.

### Code for reproduction

```python
import torch
import torchio

subject = torchio.Subject(label=torchio.LabelMap(tensor=torch.Tensor([[[[3, 5, 9, 15, 16, 23]]]])))
transform = torchio.SequentialLabels()
subject = transform(subject)
print(subject["label"][torchio.DATA])
```

### Actual outcome

tensor([[[[3., 1., 2., 3., 4., 5.]]]])

### Error messages

_No response_

### Expected outcome

tensor([[[[1., 2., 3., 4., 5., 6.]]]])

### System info

```shell
Platform:   Windows-10-10.0.19044-SP0
TorchIO:    0.18.74
PyTorch:    1.10.2+cu102
SimpleITK:  2.1.1 (ITK 5.2)
NumPy:      1.22.2
Python:     3.8.10 (tags/v3.8.10:3d8993a, May  3 2021, 11:48:03) [MSC v.1928 64 bit (AMD64)]
```

iamSmallY:
Fixes #840.

**Description**
The current mapping will map (3, 5, 9, 15, 16, 23) to {5: 1, 9: 2, 15: 3, 16: 4, 23: 5}, which is inconsistent with the documentation. And it can pass the test because in the testing procedure, when the label is generated, a 0 label will be added to it, and the 0 is the minimum in the label. But when we use an image as a label, if the minimum number is not 0, the `SequentialLabels` will cause a fatal bug.

I have no idea whether this is a non-breaking change, Only when the minimum number in label is 0 can lead to the same behavior with the edited `SequentialLabels`, but usually the minimum number in label is 0.

**Checklist**

### DIFF
--- a/torchio/transforms/preprocessing/label/sequential_labels.py
+++ b/torchio/transforms/preprocessing/label/sequential_labels.py
@@ -8,9 +8,9 @@ from .label_transform import LabelTransform
 class SequentialLabels(LabelTransform):
     r"""Remap the integer IDs of labels in a LabelMap to be sequential.
 
-    For example, if a label map has 6 labels with IDs (3, 5, 9, 15, 16, 23),
+    For example, if a label map has 7 labels with IDs (0, 3, 5, 9, 15, 16, 23),
     then this will apply a :class:`~torchio.RemapLabels` transform with
-    ``remapping={3: 1, 5: 2, 9: 3, 15: 4, 16: 5, 23: 6}``.
+    ``remapping={0: 0, 3: 1, 5: 2, 9: 3, 15: 4, 16: 5, 23: 6}``.
     This transformation is always `fully invertible <invertibility>`_.
 
     Args:
@@ -31,7 +31,7 @@ class SequentialLabels(LabelTransform):
             unique_labels = torch.unique(image.data)
             remapping = {
                 unique_labels[i].item(): i
-                for i in range(1, len(unique_labels))
+                for i in range(0, len(unique_labels))
             }
             transform = RemapLabels(
                 remapping=remapping,


romainVala:
Thanks for the PR
It looks good to me !

fepegar:
Hi, @iamSmallY. Thanks for your contribution!

@efirdc, could you please take a look at this PR?

fepegar:
As we have no news, I'm going to merge this but I might also refactor the label transforms so the label `0` is not handled more carefully.

fepegar:
@all-contributors please add @iamSmallY for bug.

fepegar:
Thanks for your contribution, @iamSmallY!

PR #844, opened 2022-03-22
Add SimpleElastix based coregistration transform

### LINKED ISSUE #428: Add high-level brain MRI transforms for preprocessing
**🚀 Feature**

Add some high-level brain MRI transforms for preprocessing.

**Motivation**

Arguably, there are three preprocessing steps performed on most multimodal neuroimaging studies:
1. Coregistration of the different modalities
2. Brain segmentation/skull-stripping
3. Registration to a reference space, typically MNI

I typically use [NiftyReg](https://github.com/KCL-BMEIS/niftyreg) for registration and [ROBEX](https://www.nitrc.org/projects/robex) for brain extraction. As discussed with @ReubenDo, other modern, robust and more Python-friendly tools that could be used for this are [SimpleElastix](https://simpleelastix.github.io/) or [ANTsPy](https://antspy.readthedocs.io/en/latest/) for registration and [HD-BET](https://github.com/MIC-DKFZ/HD-BET) for brain extraction, similar to what's done in [MRIPreprocessor](https://github.com/ReubenDo/MRIPreprocessor).

A related transform that might be useful is a defacing transform. The simplest way to do this is using a mask defined in the MNI space, therefore this is very related to the MNI registration.

**Pitch**

Add new preprocessing transforms, not meant to be used during training. Some ideas: `BrainExtraction`, `ToMNI`, `Coregister`. The requirements would be optional, rising an error with an informative message asking the user to install them using `pip`.

@romainVala, do you think this would be useful?

GReguig:
Fixes #428.

**Description**
I propose an implementation of a coregistration transform using [SimpleElastix](https://simpleelastix.readthedocs.io/). This adds an extra dependency to the SimpleElastix library which is an extension of SimpleITK, if I correctly understand.

**Checklist**


romainVala:
@fepegar what do you think of this PR ?

Here we have an issue : the test are failling because are falling because we do not know how to install Elastix (automatically for the testing environement)

Any idea how to do that ?

Is this dependency a problem to incorporate this PR into torchio ?

fepegar:
Hi, both. I had missed this PR. it is indeed problematic that SimpleElastix can't be simply pip-installed. I'm not sure how to solve that easily. Moreover, I'm still not convinced that TorchIO needs registration.

romainVala:
too bad if we can not add this transform just because of an  installation issue .. may be there is a way to make it optional ? (and up to the user to do the install) ...

I already discussed the motivation but let's resume it (and  we have now a publication (in review) to explain ). 
We add (and use) this coregistration in conjunction with the motion transform. We show in our work https://arxiv.org/pdf/2204.03522.pdf that there is no theoretical nor analytical solution to remove the global displacement induce by transient motion simulation.  So our conclusion is that we need to add a coregistration step (figure 5 is a good example of what happen without registration : a motion severity based on a voxel by voxel metric difference may be totally wrong)

Our next PR will be to propose this motion simulation transform (https://github.com/romainVala/torchio/blob/master/torchio/transforms/augmentation/intensity/random_motion_from_time_course.py) which propose this extra coregistration step (as an option). 

Note that this point is important for the objective of quantification of the motion artefact severity (with voxel by voxel difference metrics), but it is also important if one use motion augmentation in a segmentation context. Because the label stayed in the original space, the global displacement will add mis-coregistration issue in top of motion artifact. So in order to learn a segmentation robust to motion we advise to add the coregistration step after the motion simulation.

fepegar:
For now, I suggest you create a new repository and add this transform and the motion one there. You could make it pip-installable if someone wants to use them. I appreciate the PR! But I do think this is a bit out of the scope of TorchIO.

romainVala:
Ok
we will change the code, so that it rely on a torchio installation in the python path (instead of forking the all torchio repository and adding the transform in top of it)
This will be then a kind of torchio pluggin transform repo

I'll open a new issue when we are ready, but just curious right now if you think It could  be then "advertising" in the torchio documentation with a section like  "user added transform"

PR #853, opened 2022-04-09
Stop running duplicate tests when opening PRs

fepegar:
Hopefully, this will avoid running tests twice every time a PR is opened.


fepegar:
It didn't seem to have worked.

PR #854, opened 2022-04-09
Stop running duplicate tests when opening PRs

fepegar:
Hopefully, this will avoid running tests twice every time a PR is opened (second try).


fepegar:
This didn't work either. I haven't found any solutions for this. Closing for now.

PR #879, opened 2022-05-15
Add support to pass label keys for dict input

### LINKED ISSUE #866: Adding better support for Transform integration with dictionaries
**🚀 Feature**
As it stands, the TorchIO `Transform` class does not perform terribly well when trying to apply transforms to dictionaries containing label maps (as in segmentation tasks). This is mainly because there is no way for TorchIO to recognise that a given entry in a data dictionary belongs to a label map and converts it (through `DataParser`) into a `ScalarImage`.

**Motivation**

Sometimes using the `Subject` and `SubjectDataset` class is not very convenient, or sometimes a pipeline is already setup using data dictionaries and changing it would be somewhat cumbersome. For this reason, it would be good if, when applying transforms (i.e. `Affine`), to both an image and a label map, one could inform the TorchIO transforms about which entry of the data dictionary corresponds to the label so that the correct interpolation (usually `nearest`) can be used to transform the image. I provide an example below:

```python
import torch
import torchio

h,w,d = 10,10,10
image = torch.rand([1,h,w,d])
# assume 3 classes here
label = torch.randint(3,[1,h,w,d])

data_dict = {"image":image,"label":label}

random_affine = torchio.transforms.RandomAffine(
        include=["image","label"],p=1)

output = random_affine(data_dict)

print(data_dict["label"].dtype,len(torch.unique(data_dict["label"])))
print(output["label"].dtype,len(torch.unique(output["label"])))
```

Output:
```
torch.int64 3
torch.float32 858
```
Ideally, the output would be the exact same for both print statements above - in either case, a label map containing 3 different classes should be both the input and the output, with *at least* an integer `dtype`. However, what is clear is that, because of the way that the `Transform` and `DataParser` classes work (they do not distinguish between label maps and images when dictionaries are concerned), this leads to an unwanted application of the wrong interpolation and type conversion of the label map (`data_dict["label"]`) into `float32`, making the labels practically useless.

**Pitch**

I would like to ask the developers to patch both the `Transform` and `DataParser` classes so that they can include an additional parameter that refers to label classes (for the sake of simplicity let us call it `label_keys`). Then, the `_get_subject_from_dict` method could be adapted such to the following:

```python
    def _get_subject_from_dict(
            data: dict,
            image_keys: List[str],
            label_keys: List[str],
            ) -> Subject:
        subject_dict = {}
        for key, value in data.items():
            if key in image_keys and key not in label_keys:
                value = ScalarImage(tensor=value)
            if key in label_keys:
                value = LabelMap(tensor=value)
            subject_dict[key] = value
        return Subject(subject_dict)
```

**Alternatives**

I think this is the simplest solution that does not require changing the TorchIO paradigm of converting dictionaries to a `Subject` class object before applying transforms.

**Additional context**

I have nothing that can classify as additional context. I hope this was clear!

**Edit:** added the word "Output:" between the code and the output above

fepegar:
Resolves #866.

**Checklist**

### DIFF
--- a/tests/transforms/test_transforms.py
+++ b/tests/transforms/test_transforms.py
@@ -322,3 +322,23 @@ class TestTransform(TorchioTestCase):
         mask = transform.get_mask_from_bounds(3 * (0, 1), tensor)
         assert mask[0, 0, 0, 0] == 1
         assert mask.sum() == 1
+
+    def test_label_keys(self):
+        size = 1, 10, 10, 10
+        image = torch.rand(size)
+        num_classes = 2  # excluding background
+        label = torch.randint(num_classes + 1, size)
+
+        data_dict = {'image': image, 'label': label}
+
+        transform = tio.RandomAffine(
+            include=['image', 'label'],
+            label_keys=['label'],
+        )
+        transformed_label = transform(data_dict)['label']
+
+        # If the image is indeed transformed as a label map, nearest neighbor
+        # interpolation is used by default and therefore no intermediate values
+        # can exist in the output
+        num_unique_values = len(torch.unique(transformed_label))
+        assert num_unique_values <= num_classes + 1
--- a/tests/transforms/test_transforms.py
+++ b/tests/transforms/test_transforms.py
@@ -324,6 +324,8 @@ class TestTransform(TorchioTestCase):
         assert mask.sum() == 1
 
     def test_label_keys(self):
+        # Adapted from the issue in which the feature was requested:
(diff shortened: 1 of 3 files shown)


fepegar:
I'm just going to go ahead and assume you'll be happy with this, @josegcpa :)

fepegar:
@all-contributors please add @josegcpa for ideas

josegcpa:
> I'm just going to go ahead and assume you'll be happy with this, @josegcpa :)

Yes! Apologies for not having seen this earlier and thank you for this, stellar work 💎

PR #900, opened 2022-06-10
Add hann window function in GridAggregator as padding_mode

### LINKED ISSUE #894: Window functions for patch aggregation
**🚀 Feature**

Add the support to window functions as overlap_mode when aggregating patches with GridAggregator

**Motivation**

Windows functions, widely used in audio encoding, are good at reducing edge effects when aggregating a batch of overlapping patches. A the moment GridAggregator only support 'crop' and 'average' as overlap mode.

**Pitch**
Have the possibility to choose as an overlapping mode a window function. I expect to write something like: `tio.GridAggregator(sampler, overlap_mode='hann')`
Different types of window function can be implemented:
- Hann window
- Triangular window
- Pyramidal window
- Kaiser-Bessel window

**Additional context**
Different window functions have been proven to be more effective than other modes to overlap patches: https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0229839

LucaLumetti:
Fixes #894.

**Description**
Added the hann window function as overlapping mode method.

**Checklist**


LucaLumetti:
@fepegar is this ok to be merged? It failed a task but it looks like it automatically fixed the relative problem.

fepegar:
This looks great! Thanks, @LucaLumetti. Here's a visualization for it:

```python
import torchio as tio
window = tio.GridAggregator._get_hann_window((100, 200, 300))
tio.ScalarImage(tensor=window[None]).plot(percentiles=None)
```

Figure_1

src/torchio/data/inference/aggregator.py
```diff
@@ -99,6 +103,26 @@ def _initialize_avgmask_tensor(self, batch: torch.Tensor) -> None:
...
+        hann_window_3d = torch.as_tensor([1])
+        # create a n-dim hann window
+        for spatial_dim, size in enumerate(patch_size):
+            window_shape = np.ones_like(patch_size)
+            window_shape[spatial_dim] = size
+            hann_window_1d = torch.hann_window(
+                size + 2,
+                periodic=False,
+            )
+            hann_window_1d = hann_window_1d[1:-1].view(*window_shape)
```
fepegar, line 117: Why do the first and last values need to be omitted?

tests/data/inference/test_aggregator.py
```diff
@@ -33,23 +35,32 @@ def aggregate(self, mode, fixture):
...
-        )).reshape(1, 1, 4, 4)
+        )
         self.aggregate('average', fixture)
 
+    def test_overlap_hann(self):
+        fixture = (
+            ( 0 / 3,  2 / 3,  4 / 3,  6 / 3),  # noqa: E201, E241
+            ( 4 / 3,  6 / 3,  8 / 3, 10 / 3),  # noqa: E201, E241
+            ( 8 / 3, 10 / 3, 12 / 3, 14 / 3),  # noqa: E201, E241
+            (12 / 3, 14 / 3, 16 / 3, 18 / 3),
```
fepegar, line 60: I've reformatted this so it -hopefully- a bit more interpretable.

How did you compute these values?

src/torchio/data/inference/aggregator.py
LucaLumetti: For a window of length >=2 (and this is assured given `torch.hann_window(size+2)`) the first and the last element of the result array will be 0 and this cause numerical problems (division by 0)

fepegar: I would suggest not discarding the zeros and adding an epsilon (maybe `torch.finfo(torch.float32).eps`) before dividing. What do you think?

tests/data/inference/test_aggregator.py
LucaLumetti: Agree with the improved interpretability.
I've computed the 2D window function and the resulting values by hand, by using the initial 1d functions given by PyTorch as a starting point.

src/torchio/data/inference/aggregator.py
LucaLumetti: I had thought about this option but I've discarded it to avoid big numbers.
From a mathematical point of view, there shouldn't be any difference at all
Is there any reason why adding an epsilon is better than removing the first and last zero?
(It is worth noticing that this epsilon must be added to the N-dim hann window and not to the 1d, as this latter will cause epsilon^N to appear in the edges)

fepegar:
Thanks a lot for your contribution, @LucaLumetti!

@all-contributors please add @LucaLumetti for code

PR #912, opened 2022-06-24
Add fftshift to fourier_transform with torch.fft

### LINKED ISSUE #911: The fourier_transform is not shifted when torch.fft is available

### Bug summary

The behavior of `torchio` when applying a spike artifact in specific `spikes_positions` depends on whether `torch.fft` is available.  This is done extensively in [misas](https://chfc-cmi.github.io/misas/mr_artifacts.html#get_spike_series). 

This issue was introduced in https://github.com/fepegar/torchio/pull/389.
The reason is that, in when `torch.fft` is available, `fftshift` is not applied (compared to the case when numpy was/is used).
https://github.com/fepegar/torchio/blob/624ed396a6da85f50b8b67cbb5582c4eccc74bb9/src/torchio/transforms/fourier.py#L8-L16  Results are still consistent as in that case in the inverse transformation `ifftshift` is also not applied https://github.com/fepegar/torchio/blob/624ed396a6da85f50b8b67cbb5582c4eccc74bb9/src/torchio/transforms/fourier.py#L19-L27

I realize, that the functionality of `RandomSpike` is not affected by this difference. I still think it is a bug, as the result of `Spike` now depends on the `torch` version.

### Code for reproduction

```python
from torchio.transforms import Spike
from PIL import Image
import numpy as np
import torch

# download image from https://github.com/chfc-cmi/misas/raw/master/example/kaggle/images/1-frame014-slice005.png
image = Image.open("1-frame014-slice005.png").convert("I")

def spikeTransform(image, intensityFactor=1, spikePosition=[.1,.1]):
    image = np.expand_dims(np.array(image), 0)
    image = image/255
    image = torch.Tensor(image)
    spikePosition = [[.0] + spikePosition]
    spike = Spike(spikePosition, intensityFactor)
    image = spike.add_artifact(image, spikePosition, intensityFactor)[0]
    image = np.array((torch.clamp(image,0,1)))
    image = image*255
    image = image.astype(np.uint8)
    image = np.moveaxis(image, 0, 1)
    image = Image.fromarray(image)
    return image

spikeTransform(image)
```

### Actual outcome

The code above (executed in a jupyter notebook) produces this image:
actual

### Error messages

_No response_

### Expected outcome

Previously, or when `torch.fft` is not available, the code above produces this image:
expected

### System info

```shell
Platform:   Linux-5.15.0-39-generic-x86_64-with-glibc2.35
TorchIO:    0.18.79
PyTorch:    1.11.0+cu102
SimpleITK:  2.1.1.2 (ITK 5.2)
NumPy:      1.23.0
Python:     3.10.5 | packaged by conda-forge | (main, Jun 14 2022, 07:06:46) [GCC 10.3.0]
```

iimog:
Fixes #911

**Description**
This makes the behavior of fourier_transform and inv_fourier_transform
consistent with earlier versions of torchio. It is also independent of
whether torch or numpy is used to perform the transform.

**Checklist**

- [X] I have read the [`CONTRIBUTING`](https://github.com/fepegar/torchio/blob/main/CONTRIBUTING.rst) docs and have a developer setup (especially important are `pre-commit`and `pytest`)
- [X] Non-breaking change (would not break existing functionality)
- [X] Integration tests passed locally by running `pytest`
- [X] This pull request is ready to be reviewed
- [X] If the PR is ready and there are multiple commits, I have [squashed them and force-pushed](https://www.w3docs.com/snippets/git/how-to-combine-multiple-commits-into-one-with-3-steps.html#force-pushing-commits-7)

### DIFF
--- a/src/torchio/transforms/fourier.py
+++ b/src/torchio/transforms/fourier.py
@@ -8,7 +8,9 @@ class FourierTransform:
     def fourier_transform(tensor: torch.Tensor) -> torch.Tensor:
         try:
             import torch.fft
-            return torch.fft.fftn(tensor)
+            transformed = torch.fft.fftn(tensor)
+            fshift = torch.fft.fftshift(transformed)
+            return fshift
         except ModuleNotFoundError:
             import torch
             transformed = np.fft.fftn(tensor)
@@ -19,7 +21,9 @@ class FourierTransform:
     def inv_fourier_transform(tensor: torch.Tensor) -> torch.Tensor:
         try:
             import torch.fft
-            return torch.fft.ifftn(tensor)
+            f_ishift = torch.fft.ifftshift(tensor)
+            img_back = torch.fft.ifftn(f_ishift)
+            return img_back
         except ModuleNotFoundError:
             import torch
             f_ishift = np.fft.ifftshift(tensor)


fepegar:
@all-contributors please add @iimog for bug.

PR #981, opened 2022-10-09
Fix number of samples per subject in queue

### LINKED ISSUE #979: Wrong number of 'samples_per_volume' sampled within one epoch

### Bug summary

Hi Fernando,

First of all, thank you very much for adding this functionality (https://github.com/fepegar/torchio/pull/795), and sorry that it took me so long to test it.

I think I found that, during the optimization, Queue is sampling the wrong number of samples per volume. See the example below.

### Code for reproduction

```python
import torchio as tio                                                           
import numpy as np                                                              
from torch.utils.data.dataloader import DataLoader                              
import torch, random                                                            
                                                                                
seed = 42                                                                       
torch.manual_seed(seed)                                                         
torch.cuda.manual_seed(seed)                                                       
np.random.seed(seed)                                                            
random.seed(seed)                                                               
                                                                                
subjects = []                                                                   
for sub_id in range(1, 11):                                                     
    params = {                                                                  
        "im": tio.ScalarImage(tensor=np.random.random((1, 320, 320, 10))),      
        "num_samples": 8,                                                       
        "info": str(sub_id)                                                     
    }                                                                           
    subjects.append(tio.Subject(**params))                                      
                                                                                
patch_size = (320, 320, 1)                                                      
batch_size = 10                                                                 
                                                                                
sd = tio.SubjectsDataset(subjects)                                              
                                                                                
sampler = tio.data.UniformSampler(patch_size=patch_size)                        
queue = tio.Queue(sd, max_length=50, shuffle_patches=True,                      
        samples_per_volume=-1,                                                  
        sampler=sampler, num_workers=8, shuffle_subjects=True)                  
                                                                                
tr_loader = DataLoader(queue, batch_size=batch_size, shuffle=False,                
        pin_memory=False, num_workers=0)                                           
                                                                                   
# One epoch of training                                                            
for patch in tr_loader:                                                            
    print(patch["info"])
```

### Actual outcome

['**6**', '2', '7', '9', '7', '2', '4', '4', '5', '4']
['2', '7', '2', '1', '7', '5', '7', '9', '7', '1']
['5', '9', '5', '4', '4', '5', '1', '9', '7', '4']
['5', '9', '2', '1', '9', '7', '2', '1', '9', '4']
['5', '2', '1', '2', '**6**', '1', '4', '9', '1', '5']
['**10**', '**10**', '5', '2', '**10**', '**10**', '8', '3', '3', '3']
['8', '**10**', '3', '5', '2', '8', '8', '8', '8', '8']
['5', '8', '2', '3', '2', '5', '3', '3', '3', '5']

### Error messages

```python-traceback
The output of the previous code is the ID of each subject in each of the 8 training iterations.

No error messages, but the output disagrees with my understanding of what should happen. Maybe my understanding is wrong; see below.
```

### Expected outcome

The code above reproduces **one epoch** of training.
Queue size is **50**, training set size is **10**, and I want to sample **8 patches** per subject; in total, one epoch = 80 patches.
Thus, the Queue will be loaded **2 times**: the first time with **50** patches (first 5 lines of the output), and the second time with **30** patches (last 3 lines of the output).

I expected that each subject is sampled 8 times. However, according to the output shown above, the subjects are sampled differently. For example, subject "6" is sampled only twice, and subject "10" is sampled 5 times (in the second load). Some subjects are sampled the right number of times in the first load of the queue (e.g., "2" and "5") but then they are resampled in the second load of the queue.

Importantly, if we change the Queue size to 80 (so, all subjects are sampled in one "load" of the queue), this problem disappears. I haven't investigated this but the problem might be the way in which the number of samples per volume is tracked in the queue.

### System info

```shell
Platform:   Linux-4.15.0-142-generic-x86_64-with-glibc2.17
TorchIO:    0.18.84
PyTorch:    1.8.1+cu101
SimpleITK:  2.2.0 (ITK 5.3)
NumPy:      1.23.3
Python:     3.8.9 (default, Apr  3 2021, 01:02:10) 
[GCC 5.4.0 20160609]
```

fepegar:
Fixes #979.

**Checklist**

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -74,6 +74,7 @@ dev =
     matplotlib
     mypy
     pandas-stubs
+    parametrize
     pre-commit
     pytest
     pytest-cov
--- a/setup.cfg
+++ b/setup.cfg
@@ -74,7 +74,7 @@ dev =
     matplotlib
     mypy
     pandas-stubs
-    parametrize
+    parameterized
     pre-commit
     pytest
     pytest-cov
--- a/src/torchio/data/queue.py
+++ b/src/torchio/data/queue.py
@@ -158,10 +158,12 @@ class Queue(Dataset):
         self.num_workers = num_workers
         self.verbose = verbose
         self._subjects_iterable = None
+        self._incomplete_subject = None
+        self._num_patches_incomplete = 0
+        self._num_sampled_subjects = 0
         if start_background:
             self._initialize_subjects_iterable()
         self.patches_list: List[Subject] = []
-        self.num_sampled_patches = 0
 
     def __len__(self):
         return self.iterations_per_epoch
@@ -171,8 +173,8 @@ class Queue(Dataset):
         if not self.patches_list:
             self._print('Patches list is empty.')
             self._fill()
+            self.patches_list.reverse()
         sample_patch = self.patches_list.pop()
-        self.num_sampled_patches += 1
         return sample_patch
 
     def __repr__(self):
@@ -181,7 +183,6 @@ class Queue(Dataset):
             f'num_subjects={self.num_subjects}',
             f'num_patches={self.num_patches}',
             f'samples_per_volume={self.samples_per_volume}',
-            f'num_sampled_patches={self.num_sampled_patches}',
             f'iterations_per_epoch={self.iterations_per_epoch}',
         ]
         attributes_string = ', '.join(attributes)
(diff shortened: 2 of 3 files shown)


fepegar:
@jmlipman could you please check out this branch and try your script?

fepegar:
@alabamagan, I think this patch also fixes #938 and therefore we will be able to close #947. Sorry it took me so long to review your PR.

jmlipman:
@fepegar That was fast! Thanks for fixing it. I tried this branch on my test code and on my original code, with different queue sizes and patches per volume parameters, and it worked.

fepegar:
The caveat is that it uses a bit more max. memory: the full queue plus the unfinished subject. But I think that is acceptable.

fepegar:
Thanks for your contribution, @jmlipman!

@all-contributors please add @jmlipman for bug

alabamagan:
> @alabamagan, I think this patch also fixes #938 and therefore we will be able to close #947. Sorry it took me so long to review your PR.

Hi, no problem, thanks for your hard work!

PR #982, opened 2022-10-09
Stop loading data when copying unloaded image

### LINKED ISSUE #974: Image get loading when __copy__ is called

### Bug summary

The image gets already loaded while that should not happen (lazy loading). This is because the __copy__ method calls self.data causing the data to be loaded

See the __copy__ of the Image class:
 ```python
def __copy__(self):
        kwargs = {
            'tensor': self.data,
            'affine': self.affine,
            'type': self.type,
            'path': self.path,
        }
            if key in PROTECTED_KEYS:
                continue
            kwargs[key] = value  # should I copy? deepcopy?
        return type(self)(**kwargs)
```

### Code for reproduction

```python
my_image = tio.LabelMap(path='/path/to/an/image.nii.gz')

print(my_image._loaded)
new_image = copy(my_image)
print(my_image._loaded)
```

### Actual outcome

The data is loaded and therefore True is printed

### Error messages

```python-traceback
no error message
```

### Expected outcome

I expect that no data is loaded and False is printed twice

### System info

```shell
Platform:   Linux-5.4.0-62-generic-x86_64-with-glibc2.27
TorchIO:    0.18.84
PyTorch:    1.10.0+cu102
SimpleITK:  2.1.1 (ITK 5.2)
NumPy:      1.22.3
Python:     3.8.12 (default, Sep 10 2021, 00:16:05) 
[GCC 7.5.0]
```

fepegar:
Fixes #974.

@dmus, can you please take a look?

### DIFF
--- a/src/torchio/constants.py
+++ b/src/torchio/constants.py
@@ -11,6 +11,7 @@ TYPE = 'type'
 STEM = 'stem'
 DATA = 'data'
 AFFINE = 'affine'
+TENSOR = 'tensor'
 
 # For aggregator
 IMAGE = 'image'
--- a/src/torchio/data/image.py
+++ b/src/torchio/data/image.py
@@ -23,6 +23,7 @@ from ..constants import INTENSITY
 from ..constants import LABEL
 from ..constants import PATH
 from ..constants import STEM
+from ..constants import TENSOR
 from ..constants import TYPE
 from ..typing import TypeData
 from ..typing import TypeDataAffine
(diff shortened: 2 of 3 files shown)


dmus:
@fepegar changes look good.

only minor comment is access to the private attribute _loaded outside the class, but not really part of this merge request

fepegar:
Do you mean in the tests? I think that's fine, right?

dmus:
yes, in the test. If python would have real support for private/protected attribute it is not even possible to access it in the test. See also comment here for example: https://stackoverflow.com/questions/1093020/unit-testing-and-checking-private-variable-value

But minor detail, not really important here I think

fepegar:
Alright, thanks for reviewing and for sharing the link. I'll leave it like this as it still is an attribute that shouldn't be needed in the API, and we're just checking that the internals of the class look as they should.

I'll merge in a minute.

fepegar:
Thanks for your contribution, @dmus!

PR #983, opened 2022-10-09
Add support to unload images

### LINKED ISSUE #940: Unload image
**🚀 Feature**

Possibility to "unload" an image to save memory

**Motivation**

When working with huge datasets, it can easily happen that one runs out of memory. To avoid this, it would be nice to have the ability to "unload" an image if it is done with processing and may only be needed a lot later again to free up the storage

**Pitch**

It would look something like this:
```python
def unload(self):
    if not self.path:
        raise Warning("Cannot unload as not path is available from where the image could be loaded again")
        return
    self[DATA] = None
    self[AFFINE] = None
    self._loaded = False
```

**Alternatives**
The current workaround is to create new images with only the path that have not yet been loaded

**Additional context**

fepegar:
Resolves #940.

**Checklist**


fepegar:
@justusschock, could you please take a look?

src/torchio/data/image.py
```diff
@@ -561,6 +561,25 @@ def load(self) -> None:
...
+    def unload(self) -> None:
+        """Unload the image from memory.
+
+        Raises:
+            RuntimeError: If the images has not been loaded yet or if no path
+                is available.
+        """
+        if not self._loaded:
+            raise RuntimeError('Image cannot be unloaded as it is not loaded')
+        if self.path is None:
```
justusschock, line 573: For this to work nicer, maybe `self.path` should be updated on write (if still None at that point)?

tests/data/test_image.py
```diff
@@ -247,3 +247,17 @@ def numpy_reader(path):
         np.save(test_path, tensor)
         image = tio.ScalarImage(test_path, reader=numpy_reader)
         image.load()
+
```
justusschock, line 250: Add a test to check the required image attributes to be none

```suggestion
    def test_image_unload(self):
        path = self.get_image_path("unload")
        image = tio.ScalarImage(path)
        image.load()
        assert image._loaded
        image.unload()
        assert not image._loaded
        assert image[tio.constants.DATA] is None
        assert image[tio.constants.AFFINE] is None
```

justusschock (approved):
Looks good already, some minor suggestions!

src/torchio/data/image.py
fepegar: Do you mean setting `self.path` inside `self.save`? It sounds kind of nice, yes. But I feel like that's a bit too much magic and potentially unexpected behavior. I'm not sure about that feature.

justusschock: yes exactly as I find it weird if I saved the model and get the warning that there is no path present. Personally, I don't find it weird if the path wasn't set before

tests/data/test_image.py
justusschock: it shouldn't as the condition is
```python
if item in (DATA, AFFINE):
            if item not in self:
                self.load()
```
and `DATA` and `AFFINE` as keys are present in image, just set to None

src/torchio/data/image.py
fepegar: That makes sense. But what if the path has been set before? I think there would be some uncertainty about the content of the `path` attribute depending on whether the image has been loaded or not. I mean: if you load image from path A and save to path B, would you set `image.path` to B too?

fepegar: Maybe that's a bad example. My worry is:

- Create image from tensor, save to path X. `image.path` is the **output** path
- Load image from path A, save to path B. `image.path` is the **input** path, whereas maybe one would expect it to be the output path

fepegar: I'm going to merge the PR so this is done, we can keep discussing it and revisit later. I hope that's ok.

fepegar:
@all-contributors please add @justusschock for ideas and review

fepegar:
Thanks again for your contribution, @justusschock!

PR #998, opened 2022-11-17
Add support to invert RescaleIntensity transform

### LINKED ISSUE #993: (Potential) Reversible RescaleIntensity Transform
**🚀 Feature**
Implement the option to revert the intensity of the RescaleIntensity Transform

**Motivation**

Would be nice to be able to revert the applied transforms to be able to display the images in the original space.

**Pitch**

In the class `RescaleIntensity` in the file `torchio.transforms.preprocessing.intensity.rescale.py`, change the function `rescale` in L.94-96 to overwrite the class attribute `self.in_min_max`:

```python
if self.in_min_max is None:
    self.in_min_max = (array.min(), array.max())
in_min, in_max = self.in_min_max
```

By including the input minimum and maximum values (if not specifically specified in the instantiation of the transform), one can easily revert the min-max transform by replacing L105-109 with the following:

```python
out_range = self.out_max - self.out_min
        if self.invert_transform:
            array -= self.out_min
            array /= out_range
            array *= in_range
            array += in_min
        else:
            array -= in_min
            array /= in_range
            array *= out_range
            array += self.out_min
```

and by providing the class attribute `invert_transform = False` in `__init__`.

**Comments**

This inverse only works as intended if one does not use the percentile approach. In order to revert the clipping, a different approach to restore the initial values would need to be pursued. 

Furthermore, my testing revealed that the inverse transform produces some loss. I utilised a MRI slice of the BraTS challenge of size `[1,1,240,240]` and the inverse produced a total difference of 0.0123 summed across all pixels (2.14e-7 per pixel on average/ double-checked with `torch.mean`). Casting to float of the input array before the forward transformation did not resolve this (most likely requires further investigation) but I assume it is due to floating point errors.

nicoloesch:
Fixes #993 .

**Description**
The RescaleIntensity Tranasform constitutes a min-max transformation, which is easily reversible. As a result, the class attribute `self.invert_transform` is added and the function `rescale` is touched slightly in order to revert the transformation.
As it would be helpful for the user to have access to the class attribute `self.in_min_max` (and the respective `in_min`, `in_max`), the respective attributes are added.

**Known Issues**

I checked the reversibility of the transform. If the resulting `torch.Tensor` after applying `transform` and `transform.inverse()` to a random input vector, a difference between input and output tensor results. I assume this originates from rounding issues between int and float. If the output tensor is kept as a `float`, the difference is `6.2028e-07` per pixel (max: `3.8147e-06` min: `3.8147e-06`) for a `4x64x64x64` tensor. If the output tensor is casted to `int`, the difference becomes `0.2465` (max: `1.` min: `0.`)

```python
in_tensor = torch.randint(256, (4, 64, 64, 64))

transform = torchio.transforms.preprocessing.RescaleIntensity(out_min_max=(-1,1))

transformed = transform(in_tensor)
reverted = transform.inverse()(transformed)

diff = a-reverted.int()  # or without casting to int

mean_ = torch.mean(diff.float())
max_ = torch.max(diff.float())
min_ = torch.min(diff.float())
```

**Checklist**

- - I had a lot of warnings (510 passed, 3846 warnings), which I assume are not originating from my Code (only minor changes) but most likely from other parts of the projects


fepegar:
Excellent! Thanks a lot for your contribution, @nicoloesch!

@all-contributors please add @nicoloesch for code

PR #1021, opened 2023-02-02
1013-Queue-support-DDPsampler

hsyang1222:
Fixes #{1013}.

**Description**
Fixed an issue where torch DistributedSampler and torchio Queue inefficiently when using DistributedDataParallel.

The conventional method uses creating patches for all subjects in the queue of all processes, then using only some and just throwing away the rest.
The proposed method is to *first divide the subject by process* in all processes, then create a patch for the divided subject, and use it all.

**Checklist**

### DIFF
--- a/src/torchio/data/queue.py
+++ b/src/torchio/data/queue.py
@@ -68,6 +68,8 @@ class Queue(Dataset):
             but training will be slower.
         sampler: A subclass of :class:`~torchio.data.sampler.PatchSampler` used
             to extract patches from the volumes.
+        subject_sampler : Set sampler to get the subject from subjects_dataset.
+            None in normal cases; use DistributedSampler when using DDP.
         num_workers: Number of subprocesses to use for data loading
             (as in :class:`torch.utils.data.DataLoader`).
             ``0`` means that the data will be loaded in the main process.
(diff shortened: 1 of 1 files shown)


fepegar:
Looks good! I will update the branch, commit some minor edits and merge.

src/torchio/data/queue.py
```diff
@@ -136,13 +142,53 @@ class Queue(Dataset):
...
+    >>> # Assume a process running on distributed node 3
+    >>> rank = 3
+    >>> patch_sampler = tio.data.UniformSampler(patch_size)
+    >>> subject = tio.datasets.Colin27()
+    >>> subjects_dataset = tio.SubjectsDataset(10 * [subject])
+    >>> subject_sampler = dist.DistributedSampler(
+    ...     subjects_dataset,
+    ...     rank=local_rank,
+    ...     shuffle=True,
+    ...     drop_last=True,
```
fepegar, line 160: Would you recommend setting this to `True`?

fepegar (commented):
Thanks! looking good. Just left a question for you.

src/torchio/data/queue.py
hsyang1222: The shuffle keyword defaults to True for dist.DistributedSampler, so you can use it as is. I think your question is about drop_last.

When training with DDP, backward is promised to be handled in sync with each other. The processes in each rank don't know which iteration of which epoch they are processing, they just check that all other processes have entered the backward step.

Without drop_last, the number of iterations a process has to process when the number of subjects is not divided and dropped will vary and become unmanageable.

For example. You have 3 GPUs and rank 0-2 using DDP. The number of subjects is 31. With drop_last=False, rank 0 is allocated 11 subjects, rank 1 is allocated 10 subjects, and rank 2 is allocated 10 subjects. Watch carefully as rank 0 patches the 11th subject, puts it into the model, and tries to backward. Since ranks 1 and 2 do not have the 11th subject, they move on to the next epoch and start learning.

So rank 0 and rank 1 and 2 are learning a little bit out of sync, which is not what we wanted. The scary part is that the program doesn't show any errors. The programmer will be surprised by the sync time out error that suddenly appears after a very long period of learning. Worse yet, the program will finish learning completely, but just perform poorly. In this case, it's really hard to find the problem.

In conclusion, I think drop_last should be set to default True.

Translated with www.DeepL.com/Translator (free version)

fepegar: Thanks for the detailed explanation, makes sense.

fepegar:
Thanks for your contribution, @hsyang1222!

@all-contributors please add @hsyang1222 for code

PR #1022, opened 2023-02-07
Fix HistogramStandardization example

vedal:
Fixes #{issue_number}.

**Description**
Currently the example isn't loading the landmarks from path. This PR fixes that.

**Checklist**

- [v] I have read the [`CONTRIBUTING`](https://github.com/fepegar/torchio/blob/main/CONTRIBUTING.rst) docs and have a developer setup (especially important are `pre-commit`and `pytest`)
- [v] Non-breaking change (would not break existing functionality)
- [v] In-line docstrings updated
- [v] This pull request is ready to be reviewed


fepegar:
Hi, @vedal. Please read the contributing guide as requested in the PR template.

According to the documentation, this transform supports paths to NumPy arrays, so this change is not needed.

vedal:
Hi @fepegar, thanks for reviewing this. I think particularly the change to add a torch.load makes the example more readable, since the variable name "t1_landmarks" suggests its not a path (such as "t1_landmarks_path" used previously). 
As such, the change might not be necessary but at least increase readiblity.

Also, its unusual to store .npy with torch.save, as done if running the example once.

fepegar:
Thanks for clarifying, that makes sense. I'll take a look later.

fepegar:
I think the idea is that the list of landmarks is saved an an `npy` file, and `pth` would be used for a pickled dict. So I think the change I would do is replace `torch.save(t1_landmarks, t1_landmarks_path)` with `np.save(t1_landmarks_path, t1_landmarks)`. What do you think?

fepegar:
@all-contributors please add @vedal for doc

PR #1029, opened 2023-02-11
Make subject parameter mandatory

### LINKED ISSUE #1020: GridSampler without subject for validation

### Problem summary

I cannot reproduce the [GridSampler example](https://torchio.readthedocs.io/patches/patch_training.html#torchio.data.GridSampler) (see below) since in the current version, [subject cannot be none](https://github.com/fepegar/torchio/blob/main/src/torchio/data/sampler/grid.py#L78):

I need a regular grid sampler to always have the same patches in my validation. Using a uniform sampler is a problem because I have high class imbalance ( with 2 classes ) so the validation is sometimes almost entirely empty.

Instead I need the validation patches to be always the same, and ideally to contain all possible patches.

### Code for reproduction

```python
import torchio as tio
sampler = tio.GridSampler(patch_size=88)
colin = tio.datasets.Colin27()
for i, patch in enumerate(sampler(colin)):
    patch.t1.save(f'patch_{i}.nii.gz')
# To figure out the number of patches beforehand:
sampler = tio.GridSampler(subject=colin, patch_size=88)
len(sampler)
```

### Actual outcome

Assertion error line 78.

### Error messages

_No response_

### Expected outcome

It should create the sampler normally.

### System info

```shell
Platform:   Linux-4.15.0-135-generic-x86_64-with-glibc2.27
TorchIO:    0.18.86
PyTorch:    1.13.1+cu117
SimpleITK:  2.2.0 (ITK 5.3)
NumPy:      1.22.3
Python:     3.8.14 (default, Sep  8 2022, 00:02:10) 
[GCC 7.5.0]
```

fepegar:
Fixes #1020.

**Description**

The example in `GridSampler` was not working, as the `subject` parameter is actually mandatory. This PR fixes the example and the behavior of the class.

**Checklist**

### DIFF
--- a/src/torchio/data/sampler/grid.py
+++ b/src/torchio/data/sampler/grid.py
@@ -24,8 +24,6 @@ class GridSampler(PatchSampler):
             of size :math:`w \times h \times d`.
             If a single number :math:`n` is provided,
             :math:`w = h = d = n`.
-            This argument is mandatory (it is a keyword argument for backward
-            compatibility).
         patch_overlap: Tuple of even integers :math:`(w_o, h_o, d_o)`
             specifying the overlap between patches for dense inference. If a
             single number :math:`n` is provided, :math:`w_o = h_o = d_o = n`.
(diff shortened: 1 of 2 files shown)


fepegar:
@all-contributors please add @arthursw for bug, doc

PR #1049, opened 2023-03-04
Enable macOS and Windows tests on 3.11

fepegar:


fepegar:
Libraries are still not compatible with Python 3.11 for Windows and macOS.

PR #1120, opened 2023-11-01
Revert invertibility of `RescaleIntensity`

fepegar:
Implement changes proposed in #1116.

@nicoloesch @romainVala @mueller-franzes

### DIFF
--- a/src/torchio/transforms/preprocessing/intensity/rescale.py
+++ b/src/torchio/transforms/preprocessing/intensity/rescale.py
@@ -5,7 +5,7 @@ import numpy as np
 import torch
 
 from ....data.subject import Subject
-from ....typing import TypeRangeFloat
+from ....typing import TypeDoubleFloat
 from .normalization_transform import NormalizationTransform
 from .normalization_transform import TypeMaskingMethod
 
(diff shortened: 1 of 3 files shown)


nicoloesch (changes requested):
The transform is in the current state (with the proposed changes) still invertible as it has the attribute `self.revert_transform` (see `transform.py` in L.403 with the method `is_invertible`). As such, I would suggest to either overwrite `is_invertible` as proposed in #1116 or removing the entire code part for `invert_transform`. I would suggest the former as we do not loose any functionality nor do we make it more complicated - we just need to update the docstring (which has already been done in #1116 in my stashes but I am awaiting the response of all people of the issue). If that is choosen, the test needs to be adapted and needs to have two checks - one that it fails if `in_min_max` is not provided and one where it actually is invertible and that the inverse is the same again.

fepegar:
You're right, this still needs some work. I'll let you choose on top of these changes or your stash.

I didn't get this bit though:

> the test needs to be adapted and needs to have two checks - one that it fails if in_min_max is not provided and one where it actually is invertible and that the inverse is the same again.

Shouldn't we get rid of those tests?

romainVala:
thanks @nicoloesch ! sound good to me

fepegar:
Given the importance of the bug, I'm going to go ahead and merge this PRs. The code is still in the repo's history if we ever needed back. Thank you all for your contribution!

fepegar:
https://github.com/all-contributors please add @nicoloesch @romainVala @mueller-franzes for bug

PR #1153, opened 2024-02-19
Fix transforms taking strings instead of sequences

### LINKED ISSUE #923: the results are not expected when using the include='str', exclude='str' arguments in the transform

### Bug summary

For example, If I have several scans in the a subject such as subject['img'] subject['imgCT'], subject['label']. 
the include and exclude augments in transform will have unexpected behaviour. 

this is because in the function: 

class Subject(dict):
    def get_images_dict(
            self,
            intensity_only=True,
            include: Optional[Sequence[str]] = None,
            exclude: Optional[Sequence[str]] = None,
            ) -> Dict[str, Image]:
        images = {}
        for image_name, image in self.items():
            if not isinstance(image, Image):
                continue
            if intensity_only and not image[TYPE] == INTENSITY:
                continue
            if include is not None and image_name not in include:
                continue
            if exclude is not None and image_name in exclude:
                continue
            images[image_name] = image
        return images

in the line  "if include is not None and image_name not in include: " ,  if image_name = 'img' and include = 'imgCT'. 
the subject['img'] will still be included in the computation for the transform.

### Code for reproduction

```python
import torch
import torchio as tio

subject = tio.Subject(
  imgCT = tio.ScalarImage(os.path.join(case_path, 'imgCT.nii.gz')),
  img = tio.ScalarImage(os.path.join(case_path, 'img.nii.gz')),
  label = tio.LabelMap(os.path.join(case_path, 'segm.nii.gz'))
)
        
subject.plot()
transform = tio.Compose([
    tio.RandomAffine(degrees=(20, 30), exclude='img'),
    tio.RandomBlur(std=(3, 4), include='imgCT'),
])
transformed.plot()
```

### Actual outcome

as described above
the subject['img'] will still be included in the computation for the transform.

### Error messages

_No response_

### Expected outcome

the subject['img'] should not be included in the computation for the transform.

### System info

_No response_

fepegar:
Resolves #923.

**Checklist**

### DIFF
--- a/src/torchio/data/image.py
+++ b/src/torchio/data/image.py
@@ -34,6 +34,7 @@ from ..typing import TypeTripletFloat
 from ..typing import TypeTripletInt
 from ..utils import get_stem
 from ..utils import guess_external_viewer
+from ..utils import is_iterable
 from .io import check_uint_to_int
 from .io import ensure_4d
 from .io import get_rotation_and_spacing_from_affine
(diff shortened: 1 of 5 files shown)


fepegar:
@all-contributors please add @zzz123xyz for bug

PR #1178, opened 2024-06-14
Set upper bound of PyTorch version

fepegar:
**Description**

This doesn't work

```python
import torch
import torchio as tio

subject = tio.datasets.Colin27()
subject.load()
subjects = 10 * [subject]

from torch.utils.data import DataLoader
loader = DataLoader(subjects, batch_size=4)
batch = next(iter(loader))
batch.__class__ is dict
```

because `batch` is an instance of `Subject`.

This happens in PyTorch >=2.3 because of
- https://github.com/pytorch/pytorch/pull/120553

This PR forces PyTorch to be <2.3, temporarily.

**Checklist**

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -47,7 +47,7 @@ install_requires =
     nibabel
     numpy>=1.15
     scipy
-    torch>=1.1
+    torch>=1.1,<2.3
     tqdm
     typer[all]
 python_requires = >=3.8, <3.13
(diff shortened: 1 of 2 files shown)


fepegar:
@justusschock, do you think you could take a look at this problem? As you worked on the logic for copying subjects. If not, that's absolutely fine, but please let me know.

fzimmermann89:
@fepegar : Maybe a stupid question, but why is this important?

`isinstance(batch,dict)` is still true, as Subject is a subclass of dict.
This seems quite logical and works as I would expect. 

(I am asking because we would like to keep using torchio with pytorch 2.3 and would be willing to fix the issue -- if I could understand why this should be an issue and what the expected behavior is)

PR #1182, opened 2024-06-24
Make transform return the same type as input

haarisr:
**Description**

Transform call currently returns a new type `TypeTransformInput`. This PR makes a generic type bound by the original `TypeTransformInput` so that type checkers know the same type is being returned.

**Checklist**


fepegar:
Thanks for your contribution, @haarisr!

fepegar:
@allcontributors please add @haarisr for code

PR #1211, opened 2024-09-23
Implement missing restore feature for RandomGhosting

fepegar:
- **Enable preview mode in Ruff**
- **Fix linting error**
- **Implement missing restore feature for RandomGhosting**

Fixes #{issue_number}.

**Description**

**Checklist**

- [ ] I have read the [`CONTRIBUTING`](https://github.com/fepegar/torchio/blob/main/CONTRIBUTING.rst) docs and have a developer setup (especially important are `pre-commit`and `pytest`)

### DIFF
--- a/docs/examples/plot_3d_to_2d.py
+++ b/docs/examples/plot_3d_to_2d.py
@@ -36,7 +36,7 @@ def plot_batch(sampler):
     loader = tio.SubjectsLoader(queue, batch_size=16)
     batch = tio.utils.get_first_item(loader)
 
-    fig, axes = plt.subplots(4, 4, figsize=(12, 10))
+    _, axes = plt.subplots(4, 4, figsize=(12, 10))
     for ax, im in zip(axes.flatten(), batch['t1']['data']):
         ax.imshow(im.squeeze(), cmap='gray')
     plt.suptitle(sampler.__class__.__name__)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -133,6 +133,7 @@ filterwarnings = [
 quote-style = 'single'
 
 [tool.ruff.lint]
+preview = true
 select = [
     "B",
     "E",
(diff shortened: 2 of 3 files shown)


fepegar:
@romainVala @sravan953 I've implemented the feature as expected, but I'm not sure it's very useful. What do you think? Here's an example:

```python
from itertools import product

import numpy as np
import torchvision.transforms as T
from PIL import Image
from torchvision.utils import make_grid

import torchio as tio

t1 = tio.datasets.FPG().t1
t1 = tio.ToCanonical()(t1)

restores = None, 0.1, 0.25, 0.5, 0.75
intensities = 0.25, 0.5, 0.75, 1

images = {}
t1_slice = t1.data[0, :, :, t1.shape[2] // 2].float()
min_data, max_data = t1_slice.data.min(), t1_slice.data.max()
for intensity, restore in product(intensities, restores):
    print(f'intensity={intensity}, restore={restore}')
    if restore is None:
        restore_ = None
    else:
        restore_ = restore, restore
    transform = tio.Ghosting(
        num_ghosts=2,
        intensity=intensity,
        restore=restore,
        axis=1,
    )
    transformed = transform(t1)
    t1_slice = transformed.data[0, :, :, transformed.shape[2] // 2].float()
    t1_slice = (t1_slice - min_data) / (max_data - min_data) * 255
    # t1_slice = (t1_slice - t1_slice.min()) / (t1_slice.max() - t1_slice.min()) * 255
    array = t1_slice.clamp(0, 255).numpy().astype('uint8')
    array = np.rot90(array)
    image = Image.fromarray(array)
    images[intensity, restore] = image

images_list = list(images.values())
images_list = [T.ToTensor()(image) for image in images_list]
grid = make_grid(images_list, nrow=len(restores))
T.ToPILImage()(grid)
```

image

romainVala:
Hi thanks Fernando for a quick fix

(and nice plot !)

It is not clar to me how "physical" in this implementation of MRI ghosting effect 
(In EPI this is the phase difference from one line to an other that cause the ghosting)

So difficult to say if (from a physical point of view) one should keep this parameter

But from a visual point of view, this restore parameter make sense (it remove the low frequency part of the ghost, ) and so it produce more variability of the artefact

So I think it is good to keep !

many thanks

fepegar:
@allcontributors please add @sravan953 for bug

Issue #1223, opened 2024-10-15
Certain Transformations disregard *args and **kwargs (including copy)


nicoloesch:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Bug summary

Certain children of the base class `torchio.Transform` do not adhere to the *args and **kwargs provided during `__init__` as they utilise already existing transformations in `apply_transform`. This mainly includes the kwarg `copy`, which is not passed to the transform utilised in `apply_transform`. However, other attributes may be required depending on the circumstances.

Examples of this are:
- `torchio.transforms.preprocessing.spatial.crop_or_pad.CropOrPad`:  L.282 and L.286 -> no instantiation of `Crop`/`Pad` with  `self.copy`
- `torchio.transforms.preprocessing.label.sequential_labels.SequentialLabels`: L. 55 -> no instantiation of `RemapLabels` with `self.copy`
- `torchio.transforms.preprocessing.spatial.ensure_shape_multiple.EnsureShapeMultiple`: L.136 -> no instantiation of `CropOrPad` with `self.copy`
- `torchio.transforms.preprocessing.spatial.resize.Resize`: L. 75 -> no instantiation of `CropOrPad` with `self.copy`

This extends to any transformation that re-utilises an already defined base transformation.

### Code for reproduction

```python
from torchio import CropOrPad
from torchio.datasets import Colin27

subject = Colin27()
transform = CropOrPad((64,64,64), copy=False)

transformed = transform(subject)
```

### Actual outcome

The underlying transform in `apply_transform` is not called with `copy=False` as intended in the Constructor. This results in the copying of the respective copying of the `Subject` despite not defined by the user.

### Expected outcome

All *args and **kwargs of any `Transform` should be passed onto each temporary/placeholder transform during `apply_transform` if it utilises an already implemented `Transform` opposed to a standalone implementation.

### System info

```shell
Platform:   Linux-6.8.0-45-generic-x86_64-with-glibc2.35
TorchIO:    0.20.1
PyTorch:    2.3.1
SimpleITK:  2.4.0 (ITK 5.4)
NumPy:      2.0.2
Python:     3.10.15 (main, Oct  3 2024, 07:27:34) [GCC 11.2.0]
```

fepegar:
Hi @nicoloesch. Can you please explain why this is important? I'm not sure I understand the practical implications, and your snippet seems to work as expected.

nicoloesch:
Hi @fepegar,

[...]

PR #1227, opened 2024-10-24
Fix memory leak by removing custom __copy__ logic

### LINKED ISSUE #1222: Memory leak with TorchIO 0.20.1

### Bug summary

A memory leak issue is observed in TorchIO version 20.0.1 during prolonged training sessions. The memory usage gradually increases over time, leading to an eventual out-of-memory error. This issue is not present in version 0.20.0, where memory usage remains stable under similar conditions. 

The following graphs show the evolution of RAM utilization during a basic DL training session, using the provided example code, with TorchIO version 0.20.0: 
torch20_0_0

Here with TorchIO 0.20.1:

torchio20_0_1

Do you have any tips for preventing this problem with Torch 0.20.1?

Thanks :)

### Code for reproduction

```python
import copy
from typing import Any
import torch
import torch.utils.data
import torchio as tio
import pytorch_lightning as pl
import monai
import psutil
import matplotlib.pyplot as plt


ram_usage = []
timestamps = []


def get_ram_usage():
    return psutil.virtual_memory().percent

# %% Lightning module
class BasicPLModule(pl.LightningModule):
    def __init__(self):
        super().__init__()
        self.net = monai.networks.nets.Unet(spatial_dims=3, in_channels=1, out_channels=1, channels=(8, 16, 32), strides=(2, 2)).float()

    def configure_optimizers(self):
        optimizer = torch.optim.Adam(self.parameters(), lr=0.0001)
        return optimizer

    def training_step(self, batch, batch_idx):
        inputs = batch['t1'][tio.DATA].float()
        targets = batch['brain'][tio.DATA].float()
        preds = self.net(inputs)
        return torch.nn.functional.mse_loss(preds, targets)

    def on_train_epoch_end(self):
        ram_usage.append(get_ram_usage())
        timestamps.append(self.current_epoch)


# %% Main program
if __name__ == '__main__':

    subject = tio.datasets.Colin27()
    # convert to instance of new subject class
    subject = tio.Subject(
        t1=subject['t1'],
        head=subject['head'],
        brain=subject['brain'],
    )
    subjects = 10 * [subject]
    transforms = tio.Compose([
        tio.ZNormalization(masking_method = tio.ZNormalization.mean),
        tio.transforms.CropOrPad(target_shape=192)
    ])
    subject_dataset = tio.SubjectsDataset(subjects, transforms)
    training_loader = tio.SubjectsLoader(subject_dataset, batch_size=1)
    reg_net = BasicPLModule()

    trainer_args = {
        'max_epochs': 200,
    }

    trainer_reg = pl.Trainer(**trainer_args)
    trainer_reg.fit(reg_net, training_loader)

    # Plotting the data
    plt.plot(timestamps, ram_usage)
    plt.title('RAM Usage Over Time')
    plt.xlabel('Time')
    plt.ylabel('RAM Usage (%)')
    plt.grid(True)
    plt.show()
```

### Actual outcome

After a long period :
Process finished with exit code 137 (interrupted by signal 9: SIGKILL)

### Error messages

_No response_

### Expected outcome

It's an example of code where the expected outcome is irrelevant, but the RAM utilization should remain relatively constant.

### System info

```shell
Platform:   Windows-10-10.0.22631-SP0
TorchIO:    0.20.1
PyTorch:    2.4.1+cu121
SimpleITK:  2.4.0 (ITK 5.4)
NumPy:      2.0.2
Python:     3.9.13 (tags/v3.9.13:6de2ca5, May 17 2022, 16:36:42) [MSC v.1929 64 bit (AMD64)]
```

nicoloesch:
Fixes #1222 .

**Description**

This change removes the custom `__copy__` logic in `torchio.Subject`. As a result, calling `copy.copy()` on a `torchio.Subject` will no longer create a mixture of shallow and deep copies. Therefore, to
- create a shallow copy, utilise `copy.copy(subject)`
- create a deep copy, utilise `copy.deepcopy(subject)`

Removing this mix of deep and shallow copies within the method for a shallow copy (`__copy__`) appears to be beneficial for the memory management in python and the garbage collector correctly freeing `torchio.Subject` and their images that are no longer referenced.

This required the change from `copy.copy` to `copy.deepcopy` in the `torchio.Transform` base class to not modify the original subject but rather create a copy of it that is modified using the transform (applied only if `copy=True` during instantiation of any transform). I assume this is the excepted behaviour based on the provided documentation and code logic.

**Checklist**

**Important Notes**

The absence of the memory leak has only been verified on my local machine as outlined in the [comment in the original issue](https://github.com/fepegar/torchio/issues/1222#issuecomment-2433980948) and **should be verified independently by the reviewer on their machine** given the instructions in the linked comment.


nicoloesch:
**Update**: I am able to run my models with roughly 6GB of memory usage instead of 70-100GB. It appears the change fixed the memory leak.

romainVala:
Not sure this is related to this issue, but I now wonder what the copy argument of transform is intended for ?
I thought that copy=False would then apply the transform directly to the input subject
but :
```
import torchio as tio
suj = tio.datasets.Colin27()
tr = tio.RandomAffine(copy=False)
sujt = tr(suj)

#even though copy=False, sujt is a new instance of suj   they are different (and have different data

suj.history
[]
sujt.history
[Affine(scales=(tensor(0.9252,  ...


```

nicoloesch:
@romainVala I assume the copy argument is to make a deep copy of the subject and only operate on a copy of the subject, whereas the initial loaded subject stays unchanged. I think one use case of `copy=True` could be if the Subject contains `Image` elements that are initialised with a tensor opposed to an image path. However, I also was wondering whether we should not default this argument to `False` instead as we would re-load the inidividual every time we iterate over the `DataLoader` if we provide a path argument -  At least that is my understanding. What are your thoughts @fepegar?

tests/data/test_subject.py
```diff
@@ -186,3 +186,30 @@ def test_subjects_batch(self):
...
+        sub_copy['t1'].set_data(new_tensor)
+
+        # The data of both copies needs to be the same as we are using a shallow copy
+        assert torch.allclose(sub_copy['t1'].data, copy_original_subj['t1'].data)
+        # The data of the original subject should not be modified
+        assert not torch.allclose(
+            sub_copy['t1'].data, self.sample_subject['t1'].data
+        ) and not torch.allclose(
+            copy_original_subj['t1'].data, self.sample_subject['t1'].data
+        )
```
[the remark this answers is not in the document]
nicoloesch: Done and commit is pushed!

fepegar:
> we would re-load the individual every time we iterate over the DataLoader if we provide a path argument

Is this not what you would expect? Otherwise, you might run out of memory if your dataset is not small.

nicoloesch:
> Is this not what you would expect? Otherwise, you might run out of memory if your dataset is not small.

This is exactly what I expect but wouldn't then the copy be entirely obsolete? We are re-loading it from disk every time so we always have the initial state available. What is the point to protect (through copying) the initial subject, if we never re-store it, i.e. change the image on disk?

romainVala:
this was exactly my point, I do not see how the copy argument change the behaviour

If I test with a tensor, it is the same

```
import torchio as tio, numpy as np, torch

img = tio.ScalarImage(tensor=torch.rand((1,10,10,10)), affine=np.eye(4))
suj = tio.Subject({'image1':img})
tr = tio.RandomAffine(copy=False)
sujt = tr(suj)

suj.image1.data[0,0,0,:4]
Out[16]: tensor([0.7858, 0.4274, 0.7144, 0.6861])

sujt.image1.data[0,0,0,:4]
Out[17]: tensor([0.7443, 0.5312, 0.6450, 0.3963])

```
So both original subject and the transformed one are identical despite copy=False argument. 
but ok I find out where does it come from

 RandomAffine create an other transform Affine, but forget to pass the copy argument 
therefore the real transform here is done with copy=True !

If I change the line 
https://github.com/fepegar/torchio/blob/cfc8aaff11f346cc3286257ab7903464e7d5cfdb/src/torchio/transforms/augmentation/spatial/random_affine.py#L187
to 
`transform = Affine(**self.add_include_exclude(arguments), copy=self.copy)`

then 
`tr = tio.RandomAffine(copy=False)`
is now working as expected (ie suj and sujt are now identical !)

so this is indeed related to this issue #1223

Note that it is now also working for my first exemple (with subject colin) so the copy argument act similary wheter or not one use tensor as input

src/torchio/datasets/slicer.py
```diff
@@ -36,7 +36,7 @@
 }
 
 
-class Slicer(_RawSubjectCopySubject):
+class Slicer(Subject):
```
justusschock, line 39: nit: this changes behavior where it's not necessary.

before a copy of this class was of type `Subject` now i's only a instance of a subclass.

It's fine by me in general, I just remember these mechanics to be required like that in the past. Thoughts here @fepegar ?

justusschock (approved):

src/torchio/datasets/slicer.py
nicoloesch: @justusschock Thanks for taking a look at the PR!

 The `_RawSubjectCopySubject` inherits from `Subject` and implements **only** the following:
```python
class _RawSubjectCopySubject(Subject):
    def __copy__(self):
        return _subject_copy_helper(self, Subject)
```

I removed the entire class as 
1. It already implemented exactly what `Subject` did, i.e. a custom `__copy__`, with the exact same logic using `_subject_copy_helper`
2. As discussed in the entire issue, the custom `__copy__` with a mix of deep and shallow copies resulted in a memory leak

Based on the logic of the custom `__copy__` method, it appears we want to protect the `Image` and ultimately the mutable tensors in each of them as we are manually creating a deep copy just for `Image` but keeping all other attributes as a shallow copy regardless of whether they are mutable or not. I figured that mixing both in the shallow copy dunder/magic method most likely resulted in the memory leak, which has been shown in #1222 . To create a deep copy we can just rely on the regular `copy.deepcopy` without the requirement of a custom logic, particularly both `Subject` and `Image` are of type Dict and just copy each of the key-value pairs. 

To check for it, I implemented two new tests for both shallow and deep copies. You will see that both new tests pass as expected. In addition, everywhere we are creating a shallow copy of the subject I changed it to a deep copy to really protect it.

If we should have `copy=True` as a default argument to `Transform` might be a different discussion as we never change a tensor (an `Image`) on disk so we never really change it permanently anyways, which might make the copy itself redundant.

romainVala: > If we should have `copy=True` as a default argument to `Transform` might be a different discussion as we never change a tensor (an `Image`) on disk so we never really change it permanently anyways, which might make the copy itself redundant.

I do not fully understand, this argument: Of course we do not want to change the file on disk. The issue is only about Subject variable in a python script. I advocate for keeping `copy=True` as default value, because most of the time when we wrote 
`subject_transformed = transform(subject)`
on expect to have 2 different instance of `Subject` (unless we explicitly ask for `copy=False`)
... no ?

nicoloesch: Well I figured that the copy of the `Subject` is in place to protect the initial state of the `Subject` prior to any modification. I am specifically refering to any `Transform`, that may alter the content of the `Subject` and more precisely the `Image`, which usually is loaded from disk (multiple nii.gz or similar medical images). However, every modification is only temporary as we **never** modify the true origin of any `Subject`, i.e. the original image on disk is never replaced in the code unless the user actively calls `Image.save`. The question/thought I had was why protect something that we never permanently change? We always reload the ground-truth from disk whenever we access any `Image`/`Subject`. So why have `copy=True` as a default, that may or may not incur a memory leak as shown in #1222 without any real purpose?

romainVala: Not sure ...
 The real purpose of copy=True is not to protect the file on disk but only to protect the loaded subject so that you can write 
```
#with transfrom being any torchio transform
subject_transformed = tranform(subject)
```
with copy=True default (in the transform) then subject_transformed and subject have now two different contents

This is the default behaviour we want to keep, and therefore I think it is important to keep copy=True by default ...

nicoloesch: Maybe I have an issue in my thinking. Could you explain to me why it is important to do a copy by default of a `Subject`, if we never save the initial subject and never return the initial subject, which means it will be just deleted by the garbage collector eventually? We are just returning the modified/transformed `Subject`, forget about the initial one and keep progressing in the code. Why not do the tensor operations in-place? Why "protect" something we do not care about anymore?

romainVala: ok, this is probably just a matter of the exact use case. I can imagine cases where we need to keep the initial state of subject's images after the transform in order to **access it latter** in the python script (without reloading it from disk). For instance if one wish to compare the image before and after a motion artefact. 
So my point was just to say that ` copy=True ` is not useless .

this being said, I am not sure if this should the it be the default ... 

Anyway we do agree, that it is important to make the transform compliant to the user choice (copy=True or False) so indeed #1223 need to be fixed .

fepegar:
@nicoloesch I'm assuming #1228 needs to be merged before this?
@romainVala are you happy with the changes in this PR?

Sorry folks, I'm a bit busy these days.

sh-shahrokhi:
Hello, Is there any chance that this get merged?

nicoloesch:
@sh-shahrokhi I am not entirely sure when this is getting merged into the main branch. In the meantime, you can still utilise my feature branch that alleviates the issue (also referenced in the PR). The only ``issue'' is that its 0.20.1 instead of the newest version of torchio

fepegar:
Picking this up now.

fepegar:
Thanks everyone for the discussion and your patience, and of course thanks @nicoloesch for your great work!

PR #1228, opened 2024-10-24
Provide init_kwargs to every transform within another transform

### LINKED ISSUE #1223: Certain Transformations disregard *args and **kwargs (including copy)

nicoloesch:
Fixes #1223.

**Description**

Certain transforms utilise already implemented transforms during their `apply_transform` method. However, these transforms are instantiated without the original arguments passed to the outer transform. These mainly include `copy` but may also extend to `include`, `exclude`, `p` etc.

This commit appends the initial arguments to each new instantiation of a base transform to reflect the user preference.

**Checklist**


fepegar (changes requested):
Thanks, @nicoloesch! Could you please just add an explanation of what `get_base_args` does/why it exists in `Transform.get_base_args`? Just so our future selves or other maintainers have some context about it. I'll approve and merge after that.

nicoloesch:
@fepegar I added a docstring description to the method. I hope it is detailed enough!

fepegar (approved):

PR #1306, opened 2025-05-31
Add support to set figure title in plot method

toufiqmusah:
**Description**

Included plt.suptitle() to the plot_volume() method in visualization.py, to enable plotting subjects with a main title.
Added two args (figtitle: str None, fig title_y:int 0.75).

[Here](https://gist.github.com/toufiqmusah/1948f5f0aa1a7c605acec17bc71ae15a) is a GitHub Gist of it working, along with tests.

No title:
image

Title:
image

**Checklist**

- Changes are
- [ ] Documentation updated


src/torchio/visualization.py
```diff
@@ -39,6 +39,8 @@ def plot_volume(
     xlabels=True,
     percentiles=(0.5, 99.5),
     figsize=None,
+    figtitle=None,
```
fepegar, line 42: I think just `title` should be fine.

src/torchio/visualization.py
```diff
@@ -39,6 +39,8 @@ def plot_volume(
...
+    figtitle_y=0.75,
```
fepegar, line 43: I'm not sure this is needed, and this value certainly doesn't generalize well, e.g. if you try with `FPG`.

src/torchio/visualization.py
```diff
@@ -102,6 +104,9 @@ def plot_volume(
     axi_axis.set_title('Axial')
 
     plt.tight_layout()
+    if figtitle is not None:
+        plt.suptitle(figtitle, y=figtitle_y, fontsize=15)
```
fepegar, line 108: Instead of hard-coding the font size, can you remove it or add a kwarg such as `title_font_size`?

fepegar (changes requested):
Thanks, @toufiqmusah.

fepegar (approved):

fepegar:
Thanks for your contribution, @toufiqmusah.

@allcontributors, please add @toufiqmusah for code

fepegar:
@allcontributors please add @toufiqmusah for code

PR #1308, opened 2025-06-04
Stop copying whole image before cropping

### LINKED ISSUE #1305: PatchSampler slowdown after upgrading

- [x] I have searched the existing issues

### Bug summary

We recently updated from torchio 0.20.3 to the most recent version and we noticed a significant slowdown in our pipeline (similar to [393](https://github.com/TorchIO-project/torchio/issues/393)). This makes our trainings take an impractically long time.

We have traced it back to the change made in [PR 1227](https://github.com/TorchIO-project/torchio/pull/1227) which causes a deepcopy to be applied for each patch to be pulled from a subject, instead of only a shallow copy as before 0.20.4.

### Code for reproduction

```python
import torchio as tio
import torch
from copy import deepcopy
from datetime import datetime

def datapipe(subjects):
    transforms = tio.Compose([
        tio.Resample((1.0, 1.0, 1.0)),  # Resample to isotropic voxel size
        tio.RandomGamma(
            log_gamma=(0.8, 1.2),  # Randomly adjust gamma
            p=0.5,  # Apply with a probability of 50%
        ),
        tio.RescaleIntensity(percentiles=(0.5, 99.5), include=['source',]),  # Normalize intensities
    ])

    # Create a patch-based data pipeline
    dataset = tio.SubjectsDataset(subjects, transform=transforms)
    queue = tio.Queue(dataset, max_length=100, samples_per_volume=50, sampler=tio.WeightedSampler((24,24,24), probability_map='sampling_map'), num_workers=4)
    loader = tio.SubjectsLoader(queue, batch_size=24)
    return loader


if __name__ == '__main__':
    # Example subject with large 3D images
    subject = tio.Subject(
        source = tio.ScalarImage(tensor=torch.rand(1, 512,512, 224)),
        sampling_map=tio.ScalarImage(tensor=torch.ones(1, 512, 512, 224)),
        target=tio.LabelMap(tensor=torch.randint(0, 2, (1, 512, 512, 224))),
    )
    # Create a list of subjects for the data pipeline
    subjects = []
    for _ in range(10):
        subjects.append(deepcopy(subject))

    dataloader = datapipe(subjects)
    dataloader = iter(dataloader)
    starttime = datetime.now()
    for _ in range(5):
        batch = next(dataloader)
        # Process the batch
        print(batch.keys())
    endtime = datetime.now()
    print("Time taken:", endtime - starttime)
```

### Actual outcome

On torchio < 0.20.4:
Time taken: 0:00:20.214979
On most recent version:
Time taken: 0:01:19.506775

### Expected outcome

I would hope there is a way to speed up the pipeline to how it was before. My proposal would be instead of having the Crop Transform copy the whole subject, we would create a new subject and only copy the patches of the volumes that we need.

### System info

```shell
New:
Platform:   Linux-5.15.0-139-generic-x86_64-with-glibc2.35
TorchIO:    0.20.8
PyTorch:    2.7.0+cpu
SimpleITK:  2.5.0 (ITK 5.4)
NumPy:      2.2.6
Python:     3.11.0rc1 (main, Aug 12 2022, 10:02:14) [GCC 11.2.0]

Old:
Platform:   Linux-5.15.0-139-generic-x86_64-with-glibc2.35
TorchIO:    0.20.3
PyTorch:    2.7.0+cpu
SimpleITK:  2.5.0 (ITK 5.4)
NumPy:      2.2.6
Python:     3.11.0rc1 (main, Aug 12 2022, 10:02:14) [GCC 11.2.0]
```

StijnvWijn:
Fixes #1305.

**Description**

A slowdown while sampling patches that are much smaller than the image in Torchio >= 0.20.4

**Checklist**

- Changes are
- [ ] Documentation updated

### DIFF
--- a/src/torchio/transforms/preprocessing/spatial/crop.py
+++ b/src/torchio/transforms/preprocessing/spatial/crop.py
@@ -57,7 +57,7 @@ class Crop(BoundsTransform):
             
             # Copy all non-image attributes
             for key, value in sample.items():
-                if value not in self.get_images(sample):
+                if key not in sample.get_images_dict(intensity_only=False, include=self.include, exclude=self.exclude).keys():
                     sample_attributes[key] = copy.deepcopy(value)
                 else:
                     sample_attributes[key] = self.crop_image(value, index_ini, index_fin)
--- a/src/torchio/transforms/preprocessing/spatial/crop.py
+++ b/src/torchio/transforms/preprocessing/spatial/crop.py
@@ -56,26 +56,30 @@ class Crop(BoundsTransform):
         index_fin = np.array(subject.spatial_shape) - high
 
         if self.copy_patch:
-            # Create a new subject with only the cropped patch
-            sample_attributes = {}
+            # Create a clean new subject to copy the images into
+            # We do this __new__ to avoid calling __init__ so we don't have to specify images immediately
+            cropped_subject = subject.__class__.__new__(subject.__class__)
             image_keys_to_crop = subject.get_images_dict(
                 intensity_only=False, include=self.include, exclude=self.exclude
             ).keys()
-            # Copy all non-image attributes
-            for key, value in subject.items():
+            keys_to_expose = subject.keys()
+            # Copy all attributes we don't want to crop
+            # __dict__ returns all attributes, instead of just the images
+            for key, value in subject.__dict__.items():
                 if key not in image_keys_to_crop:
-                    sample_attributes[key] = deepcopy(value)
+                    copied_value = deepcopy(value)
+                    # Setting __dict__ does not allow key indexing the attribute
+                    # so we set it explicitly if we want to expose it
+                    if key in keys_to_expose:
+                        cropped_subject[key] = copied_value
+                    cropped_subject.__dict__[str(key)] = copied_value
                 else:
-                    sample_attributes[key] = self._crop_image(
-                        value, index_ini, index_fin
-                    )
-            cropped_sample = type(subject)(**sample_attributes)
+                    # Images are always exposed, so we don't worry about setting __dict__
+                    cropped_subject[key] = self._crop_image(value, index_ini, index_fin)
 
-            # Copy applied transforms history
-            cropped_sample.applied_transforms = deepcopy(subject.applied_transforms)
-
-            cropped_sample.update_attributes()
-            return cropped_sample
+            # Update the __dict__ attribute to include the cropped images
+            cropped_subject.update_attributes()
+            return cropped_subject
         else:
             # Crop in place
             for image in self.get_images(subject):
--- a/src/torchio/transforms/preprocessing/spatial/crop.py
+++ b/src/torchio/transforms/preprocessing/spatial/crop.py
@@ -1,7 +1,9 @@
+import copy
 import nibabel as nib
 import numpy as np
 
 from ....data.subject import Subject
+from ....data.image import Image
 from .bounds_transform import BoundsTransform
 from .bounds_transform import TypeBounds
 
@@ -24,6 +26,11 @@ class Crop(BoundsTransform):
             If only one value :math:`n` is provided, then
             :math:`w_{ini} = w_{fin} = h_{ini} = h_{fin}
             = d_{ini} = d_{fin} = n`.
+        copy: bool, optional
+            This transform overwrites the copy argument of the base transform and 
+            copies only the cropped patch, instead of the whole image.
+            If ``True``, the cropped image will be copied to a new subject.
+            If ``False``, the patch will be cropped in place. Default: ``True``.
         **kwargs: See :class:`~torchio.transforms.Transform` for additional
             keyword arguments.
 
@@ -31,26 +38,64 @@ class Crop(BoundsTransform):
         :class:`~torchio.transforms.CropOrPad` instead.
     """
 
-    def __init__(self, cropping: TypeBounds, **kwargs):
-        super().__init__(cropping, **kwargs)
+    def __init__(self, cropping: TypeBounds, copy=True, **kwargs):
+        self.copy_patch = copy
+        super().__init__(cropping, copy=False, **kwargs)
         self.cropping = cropping
         self.args_names = ['cropping']
 
-    def apply_transform(self, sample) -> Subject:
+    def apply_transform(self, sample: Subject) -> Subject:
         assert self.bounds_parameters is not None
         low = self.bounds_parameters[::2]
         high = self.bounds_parameters[1::2]
         index_ini = low
         index_fin = np.array(sample.spatial_shape) - high
-        for image in self.get_images(sample):
-            new_origin = nib.affines.apply_affine(image.affine, index_ini)
-            new_affine = image.affine.copy()
-            new_affine[:3, 3] = new_origin
-            i0, j0, k0 = index_ini
-            i1, j1, k1 = index_fin
+        
+        if self.copy_patch:
+            # Create a new subject with only the cropped patch
+            sample_attributes = {}
+            
+            # Copy all non-image attributes
+            for key, value in sample.items():
+                if value not in self.get_images(sample):
+                    sample_attributes[key] = copy.deepcopy(value)
+                else:
+                    sample_attributes[key] = self.crop_image(value, index_ini, index_fin)
+            cropped_sample = type(sample)(**sample_attributes)
+            
+            # Copy applied transforms history
+            cropped_sample.applied_transforms = copy.deepcopy(sample.applied_transforms)
+            
+            cropped_sample.update_attributes()
+            return cropped_sample
+        else:
+            # Crop in place
+            for image in self.get_images(sample):
+                self.crop_image(image, index_ini, index_fin)
+            return sample
+    
+    def crop_image(self, image: Image, index_ini: tuple, index_fin: tuple) -> None:
+        new_origin = nib.affines.apply_affine(image.affine, index_ini)
+        new_affine = image.affine.copy()
+        new_affine[:3, 3] = new_origin
+        i0, j0, k0 = index_ini
+        i1, j1, k1 = index_fin
+        
+        # Crop the image data
+        if self.copy_patch:
+            # Create a new image with the cropped data
+            cropped_data = image.data[:, i0:i1, j0:j1, k0:k1].clone()
+            new_image = type(image)(
+                tensor=cropped_data,
+                affine=new_affine,
+                type=image.type,
+                path=image.path,
+            )
+            return new_image
+        else:
             image.set_data(image.data[:, i0:i1, j0:j1, k0:k1].clone())
             image.affine = new_affine
-        return sample
+            return image
 
     def inverse(self):
         from .pad import Pad
--- a/src/torchio/transforms/preprocessing/spatial/crop.py
+++ b/src/torchio/transforms/preprocessing/spatial/crop.py
@@ -1,7 +1,7 @@
 from copy import deepcopy
 
-import nibabel as nib
 import numpy as np
+from nibabel.affines import apply_affine
 
 from ....data.image import Image
 from ....data.subject import Subject
@@ -27,12 +27,11 @@ class Crop(BoundsTransform):
             If only one value :math:`n` is provided, then
             :math:`w_{ini} = w_{fin} = h_{ini} = h_{fin}
             = d_{ini} = d_{fin} = n`.
-        copy: bool, optional
-            This transform overwrites the copy argument of the base transform and
-            copies only the cropped patch, instead of the whole image.
-            This can provide a significant speedup when cropping small patches from large images
-            If ``True``, each image will be cropped and the patch copied to a new subject.
-            If ``False``, each image will be cropped in place. Default: ``True``.
+        copy: If ``True``, each image will be cropped and the patch copied to a new
+            subject. If ``False``, each image will be cropped in place. This transform
+            overwrites the copy argument of the base transform and copies only the
+            cropped patch instead of the whole image. This can provide a significant
+            speedup when cropping small patches from large images.
         **kwargs: See :class:`~torchio.transforms.Transform` for additional
             keyword arguments.
 
@@ -41,7 +40,7 @@ class Crop(BoundsTransform):
     """
 
     def __init__(self, cropping: TypeBounds, copy=True, **kwargs):
-        self.copy_patch = copy
+        self._copy_patch = copy
         # Transform base class deepcopies whole subject by default
         # We want to copy only the cropped patch, so we overwrite the functionality
         super().__init__(cropping, copy=False, **kwargs)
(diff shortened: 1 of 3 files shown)


fepegar:
@nicoloesch could you please comment on the general design? I can then fix formatting etc.

src/torchio/transforms/preprocessing/spatial/crop.py
```diff
@@ -24,33 +27,83 @@ class Crop(BoundsTransform):
...
+
+            cropped_sample.update_attributes()
+            return cropped_sample
+        else:
+            # Crop in place
+            for image in self.get_images(sample):
+                self.crop_image(image, index_ini, index_fin)
+            return sample
+
+    def crop_image(self, image: Image, index_ini: tuple, index_fin: tuple) -> None:
```
nicoloesch, line 85: I think we could also make this a @staticmethod and just pass the copy attribute down as another function arg as we do not need to have access to the state of self beyond this one flag. But totally optional. Same with whether we want to expose it using `crop_image` or make it more internal with `_crop_image`.

src/torchio/transforms/preprocessing/spatial/crop.py
```diff
@@ -24,33 +27,83 @@ class Crop(BoundsTransform):
...
-            i0, j0, k0 = index_ini
-            i1, j1, k1 = index_fin
+
+        if self.copy_patch:
+            # Create a new subject with only the cropped patch
+            sample_attributes = {}
+
+            # Copy all non-image attributes
+            for key, value in sample.items():
+                if (
```
nicoloesch, line 61: This gets the images_dict for every iteration. Maybe we could set it before the loop as a variable (e.g. imaging_keys) that goes once over the images, stores the keys as thats all we need and then we just need to check if the key is in it or not. But doubt that provides a lot of speed-up but might be a bit cleaner

src/torchio/transforms/preprocessing/spatial/crop.py
```diff
@@ -24,33 +27,83 @@ class Crop(BoundsTransform):
...
-    def __init__(self, cropping: TypeBounds, **kwargs):
-        super().__init__(cropping, **kwargs)
+    def __init__(self, cropping: TypeBounds, copy=True, **kwargs):
+        self.copy_patch = copy
+        super().__init__(cropping, copy=False, **kwargs)
         self.cropping = cropping
         self.args_names = ['cropping']
 
-    def apply_transform(self, sample) -> Subject:
+    def apply_transform(self, sample: Subject) -> Subject:
```
nicoloesch, line 48: Maybe have it named as subject instead of sample to make it even clearer. Same with cropped_sample to cropped_subject. I am aware this was called `sample` prior to your change but it might confuse a bit

src/torchio/transforms/preprocessing/spatial/crop.py
```diff
@@ -1,6 +1,9 @@
+import copy
```
nicoloesch, line 1: I do not know how python handles this exaclty and it may work out this way but we are using the module `copy` and a flag `copy` with the exact same name. This may bear the risk of random behaviour with overwriting of attributes or something (again I am not entirely sure). Maybe it is better to use `from copy import deepcopy` as thats the only one we need and that alleviates any potential risk.

src/torchio/transforms/preprocessing/spatial/crop.py
```diff
@@ -24,33 +27,83 @@ class Crop(BoundsTransform):
             If only one value :math:`n` is provided, then
             :math:`w_{ini} = w_{fin} = h_{ini} = h_{fin}
             = d_{ini} = d_{fin} = n`.
+        copy: bool, optional
```
nicoloesch, line 30: The description but could be a bit more consistent. This is just an idea:
- If ``True``, each image will be cropped and the patch copied to a new subject.
- If ``False``,  *each image* will be cropped in place.

Maybe also a justification as to why the default mechanism was circumvented, i.e. to provide a speed-up in case with relative small crops compared to larger images, which prevents the entire image from being copied.

src/torchio/transforms/preprocessing/spatial/crop.py
```diff
@@ -24,33 +27,83 @@ class Crop(BoundsTransform):
...
 
     .. seealso:: If you want to pass the output shape instead, please use
         :class:`~torchio.transforms.CropOrPad` instead.
     """
 
-    def __init__(self, cropping: TypeBounds, **kwargs):
-        super().__init__(cropping, **kwargs)
+    def __init__(self, cropping: TypeBounds, copy=True, **kwargs):
+        self.copy_patch = copy
+        super().__init__(cropping, copy=False, **kwargs)
```
nicoloesch, line 44: ***Maybe*** leave an inline comment here to say that this circumvents the copying mechanism in case it is not entirely clear in the description/docstring above. Could also just be included in the docstring with wording about that the base mechanism of copying is handled internally in `apply_transform` instead of in the base class.

nicoloesch (commented):
Really like the changes @StijnvWijn and this is exactly how I would have done it. Some changes are purely stylistic, others require some feedback from @fepegar about formatting and docstring description to make it as clear as possible.
Most of my comments are suggestions instead of requirements.

src/torchio/transforms/preprocessing/spatial/crop.py
StijnvWijn: Hmm good point and we only need the deepcopy, so I changed it

src/torchio/transforms/preprocessing/spatial/crop.py
StijnvWijn: Hmm yeah, I didn't want to change it since technically it might be something else if I read the Transform docs, but I am pretty sure it didn't work with anything except subjects before either.

src/torchio/transforms/preprocessing/spatial/crop.py
StijnvWijn: I am not sure what the advantage would be to making it a staticmethod, but we can definitely make it an internal function.

StijnvWijn:
@nicoloesch Do you have any comments on the new copy mechanism?

nicoloesch (approved):
All my comments have been addressed! Thank you @StijnvWijn

src/torchio/transforms/preprocessing/spatial/crop.py
fepegar: > Hmm yeah, I didn't want to change it since technically it might be something else if I read the Transform docs, but I am pretty sure it didn't work with anything except subjects before either.

IIRC this was `sample` for historical reasons and by the time the input reaches `apply_transform` it's always an instance of `Subject`, so I agree `subject` is better here.

fepegar:
Thank you both for your contribution!

fepegar:
> Hmm that is very sad.

I know, and I'm sorry I had to revert this. I'm sure we'll be able to find a solution! I'm happy to refactor the design if needed.

> I am not sure where the issue would occur, because all attributes seem to be deepcopied from the original input. Do you have any idea where to start?

I would start with the snippet I shared above.

StijnvWijn:
Thanks for you quick response!

I am not sure I can reproduce your issue. In your snippet, indeed the memory usage increases slightly between the instantiation of the dataset and the creation of the cropped patches, but IMO this is expected, because we make a copy of the subject with all the old attributes and put a copy of the cropped area of the original image into the subject.

Also if I run the subject generation loop multiple times, I see some minor fluctuations in memory usage, but they go both up and down between loops, so not the behaviour I would expect with a memory leak. 

I am running it on a server with quite a bit of RAM, just like I described in my original issue, so that might also have an impact on my results.

StijnvWijn:
@fepegar 

I would not mind to work on this, but I can't test whether my implementation works currently. Shall I make a new issue or PR to discuss this issue or continue it here?

nicoloesch:
I may also be able to assist but I do have currently very limited time as I am in the final stages to submit my PhD. I may have some time this week and will start with the snippet you, @fepegar, provided and whether I can reproduce the memory leak!

StijnvWijn:
Ah awesome, that is great news for you. Good luck!

It is not too urgent as we are currently just using an older version of tio, so you don't need to worry too much about it :)

fepegar:
Thanks for trying. When I ran it, my memory was increasing much faster
(e.g. 4 GB after a few iterations). I’ll try to replicate and share my
results here.

StijnvWijn:
Thanks for your quick response!

I'm sorry, apparently it does not show up on the server, but if I run it on my laptop, I can replicate your issue. I see that the pytorch versions are different, so maybe its related to that? There were some changes related to the default collation that made us stay with an older Pytorch version, but I will check.
Laptop:
```
Platform:   Linux-6.14.0-24-generic-x86_64-with-glibc2.41
TorchIO:    0.20.16
PyTorch:    2.7.0+cu126
SimpleITK:  2.5.0 (ITK 5.4)
NumPy:      2.3.1
Python:     3.11.12 (main, Apr  9 2025, 15:24:58) [GCC 14.2.0]
```
torchio 0.20.8:
| Laptop                                    | Server                                    |
|-------------------------------------------|-------------------------------------------|
| Checking memory usage before importing... | Checking memory usage before importing... |
| Memory usage: 12.73 MB                    | Memory usage: 12.08 MB                    |
| Importing...                              | Importing...                              |
| Memory usage: 549.42 MB                   | torchio version 0.20.8                    |
| Creating subjects list...                 | Memory usage: 325.84 MB                   |
| Memory usage: 549.93 MB                   | Creating subjects list...                 |
| Instantiating SubjectsDataset...          | Memory usage: 325.84 MB                   |
| Memory usage: 549.93 MB                   | Instantiating SubjectsDataset...          |
| Loading...                                | Memory usage: 325.84 MB                   |
| Memory usage: 559.93 MB                   | Loading...                                |
| Memory usage: 560.28 MB                   | Memory usage: 335.29 MB                   |
| Memory usage: 560.19 MB                   | Memory usage: 335.18 MB                   |
| Memory usage: 560.70 MB                   | Memory usage: 335.02 MB                   |
| Memory usage: 559.93 MB                   | Memory usage: 334.73 MB                   |
| Memory usage: 560.55 MB                   | Memory usage: 334.42 MB                   |
| Memory usage: 560.43 MB                   | Memory usage: 335.80 MB                   |
| Memory usage: 560.38 MB                   | Memory usage: 334.57 MB                   |
| Memory usage: 559.98 MB                   | Memory usage: 336.84 MB                   |
| Memory usage: 560.30 MB                   | Memory usage: 335.12 MB                   |
|                                                                    | Memory usage: 336.25 MB                   |

Torchio 0.20.10: 
| Laptop                                    | Server                                    |
|-------------------------------------------|-------------------------------------------|
| Checking memory usage before importing... | Checking memory usage before importing... |
| Memory usage: 12.61 MB                    | Memory usage: 12.02 MB                    |
| Importing...                              | Importing...                              |
| Torchio version: 0.20.10                  | torchio version 0.20.10                   |
| Memory usage: 550.13 MB                   | Memory usage: 325.61 MB                   |
| Creating subjects list...                 | Creating subjects list...                 |
| Memory usage: 550.64 MB                   | Memory usage: 325.61 MB                   |
| Instantiating SubjectsDataset...          | Instantiating SubjectsDataset...          |
| Memory usage: 550.64 MB                   | Memory usage: 325.61 MB                   |
| Loading...                                | Loading...                                |
| Memory usage: 1970.80 MB                  | Memory usage: 334.90 MB                   |
| Memory usage: 3381.06 MB                  | Memory usage: 334.96 MB                   |
| Memory usage: 3380.93 MB                  | Memory usage: 335.95 MB                   |
| Memory usage: 4791.33 MB                  | Memory usage: 336.54 MB                   |
| Memory usage: 4791.88 MB                  | Memory usage: 336.51 MB                   |
| Memory usage: 6201.87 MB                  | Memory usage: 336.25 MB                   |
| Memory usage: 7611.93 MB                  | Memory usage: 335.81 MB                   |
| Memory usage: 4791.07 MB                  | Memory usage: 336.03 MB                   |
| Memory usage: 6201.60 MB                  | Memory usage: 336.52 MB                   |
| Memory usage: 4791.62 MB                  | Memory usage: 336.50 MB                   |

torchio 0.20.16:
| Laptop                                    | Server                                    |
|-------------------------------------------|-------------------------------------------|
| Checking memory usage before importing... | Checking memory usage before importing... |
| Memory usage: 12.90 MB                    | Memory usage: 12.02 MB                    |
| Importing...                              | Importing...                              |
| Torchio version: 0.20.16                  | torchio version 0.20.16                   |
| Memory usage: 548.96 MB                   | Memory usage: 325.77 MB                   |
| Creating subjects list...                 | Creating subjects list...                 |
| Memory usage: 549.48 MB                   | Memory usage: 325.77 MB                   |
| Instantiating SubjectsDataset...          | Instantiating SubjectsDataset...          |
| Memory usage: 549.48 MB                   | Memory usage: 325.77 MB                   |
| Loading...                                | Loading...                                |
| Memory usage: 1969.89 MB                  | Memory usage: 335.65 MB                   |
| Memory usage: 1969.67 MB                  | Memory usage: 334.47 MB                   |
| Memory usage: 3380.48 MB                  | Memory usage: 335.05 MB                   |
| Memory usage: 1970.29 MB                  | Memory usage: 335.25 MB                   |
| Memory usage: 3380.22 MB                  | Memory usage: 335.32 MB                   |
| Memory usage: 4790.86 MB                  | Memory usage: 335.53 MB                   |
| Memory usage: 3380.21 MB                  | Memory usage: 336.39 MB                   |
| Memory usage: 4790.28 MB                  | Memory usage: 336.33 MB                   |
| Memory usage: 3380.75 MB                  | Memory usage: 336.06 MB                   |
| Memory usage: 4790.40 MB                  | Memory usage: 335.98 MB                   |

fepegar:
Interesting that it's different! But we definitely should check the PyTorch versions. However I'm not sure this would be about the collating function, because note in the script I'm not using a `DataLoader`!

StijnvWijn:
I have found an issue that might be related: If I create a new tio.Image inside of a loop, it seems to not be garbage collected. Not sure why yet, but it also causes a linear increase in memory consumption if I run the following snippet:
```
for _ in range(10):
    new_tensor = torch.ones((1, 100,100,100), dtype=torch.float32)
    new_image = tio.Image(tensor = new_tensor)
    print(f'Created tensor with shape {new_tensor.shape} and id {id(new_tensor)}')
    print_memory_usage()
```
It does not happen if the new_image is just a tensor, an ordinary dict like {'tensor': new_tensor} or a much simpler class like below, so I am not yet sure where the issue lies...
```
class TestImage(dict):
    def __init__(self, tensor, **kwargs):
        
        self['data'] = tensor
        self['affine'] = None  # Placeholder for affine, if needed
        super().__init__(**kwargs)
```

StijnvWijn:
Awesome!

No problem, It also took me quite some time to find the issue, did not even know this was how that warnings module worked haha.

Issue #48, opened 2020-01-15
Add antialiasing option to Resample and RandomAffine


fepegar:
As in https://github.com/NifTK/NiftyNet/commit/a91644b881192fb68e7faabc2065129f09d5bdd5#diff-5b79901b7b2f56394aca30c562e9e186

sairampillai:
Hey @fepegar. I would like to contribute. Would this be a good issue to start with? Any other details regarding this issue?

fepegar:
Hi @sairampillai. Thanks for your interest!
Yes, this might be a good one. I had in mind something like what I did for NiftyNet:

https://github.com/NifTK/NiftyNet/blob/7421b013e5d7852f2eab1861de0bc768995ec82b/niftynet/layer/rand_spatial_scaling.py#L67-L73

You can start by
1. Fork this repo
2. Create a branch `48-add-antialiasing` on the fork
3. Commit your changes to that branch
4. Submit a pull request to this repo, even if it's not ready

I'll review the commits and merge your changes.

Thank you!

PR #1314, opened 2025-06-18
Add antialiasing option for downsampling

fepegar:
Related to #48.

**Description**

```python
import torchio as tio

image = tio.datasets.Colin27().t1
image.plot()

factor = 8
downsampled_old = tio.Resample(factor, antialias=False)(image)
downsampled_old.plot()
downsampled_new = tio.Resample(factor, antialias=True)(image)
downsampled_new.plot()
```

image
image
image

Note e.g. the ventricles and basal ganglia in the sagittal view.

One more example with a test image. Note aliasing is smaller when enabling the new setting.

image
image
image

**Checklist**

- Changes are
- [ ] Documentation updated

### DIFF
--- a/src/torchio/transforms/preprocessing/spatial/resample.py
+++ b/src/torchio/transforms/preprocessing/spatial/resample.py
@@ -52,13 +52,15 @@ class Resample(SpatialTransform):
         label_interpolation: See :ref:`Interpolation`.
         scalars_only: Apply only to instances of :class:`~torchio.ScalarImage`.
             Used internally by :class:`~torchio.transforms.RandomAnisotropy`.
-        antialias: If ``True``, apply a Gaussian smoothing before
-            downsampling, along any dimension that will be downsampled.
-            This is useful to avoid aliasing artifacts when downsampling
-            images. The standard deviation of the Gaussian kernel
-            is computed according to the method described in Cardoso et al.,
-            `Scale factor point spread function matching: beyond aliasing in
-            image resampling
+        antialias: If ``True``, apply Gaussian smoothing before
+            downsampling along any dimension that will be downsampled. For example,
+            if the input image has spacing (0.5, 0.5, 4) and the target
+            spacing is (1, 1, 1), the image will be smoothed along the first two
+            dimensions before resampling. Label maps are not smoothed.
+            The standard deviations of the Gaussian kernels are computed according to
+            the method described in Cardoso et al.,
+            `Scale factor point spread function matching: beyond aliasing in image
+            resampling
             <https://link.springer.com/chapter/10.1007/978-3-319-24571-3_81>`_,
             MICCAI 2015.
         **kwargs: See :class:`~torchio.transforms.Transform` for additional
@@ -67,8 +69,8 @@ class Resample(SpatialTransform):
     Example:
         >>> import torch
         >>> import torchio as tio
-        >>> transform = tio.Resample(1)                     # resample all images to 1mm iso
-        >>> transform = tio.Resample((2, 2, 2))             # resample all images to 2mm iso
+        >>> transform = tio.Resample()                      # resample all images to 1mm isotropic
+        >>> transform = tio.Resample(2)                     # resample all images to 2mm isotropic
         >>> transform = tio.Resample('t1')                  # resample all images to 't1' image space
         >>> # Example: using a precomputed transform to MNI space
         >>> ref_path = tio.datasets.Colin27().t1.path  # this image is in the MNI space, so we can use it as reference/target
@@ -76,6 +78,11 @@ class Resample(SpatialTransform):
         >>> image = tio.ScalarImage(tensor=torch.rand(1, 256, 256, 180), to_mni=affine_matrix)  # 'to_mni' is an arbitrary name
         >>> transform = tio.Resample(colin.t1.path, pre_affine_name='to_mni')  # nearest neighbor interpolation is used for label maps
         >>> transformed = transform(image)  # "image" is now in the MNI space
+
+    .. note::
+        The ``antialias`` option is recommended when large (e.g. > 2×) downsampling
+        factors are expected, particularly for offline (before training) preprocessing,
+        when run times are not a concern.
 
     .. plot::
 


fepegar:
@romainVala, could you please take a look?

romainVala:
ok my multiple resample is working only because it use 3 successiv interpolation. Indeed if you compare with the same 3 consecutive resampling, but with `image_interpolation='nearest'`, then the result is crapy
So ok those 3 successive interpolation may be equivalent to gaussian filter you made for the anti-aliasing

What I do not understand, is why adding a smoothing when a simple average pooling is performing the job exactly (ie without interpolaiton ) ... ?

fepegar:
Thank you, @romainVala.

For reference, here are the images generated by your code.

image

image

image

Also added the one with nearest neighbor interpolation.

image

This is definitely a digital signal processing (DSP) problem, but I can maybe try to explain intuitively and from a DSP point of view (although my undergrad happened a long time ago).

Let's define some downsampling functions for 1D signals.

```python
>>> def avg_pool(x, k=2):
...     return torch.nn.functional.avg_pool1d(x.unsqueeze(0).unsqueeze(0), kernel_size=k, stride=k)[0, 0]
...
>>>
>>> def resample(x, f=2, mode='linear'):
...     align_corners = False if mode == 'linear' else None
...     return torch.nn.functional.interpolate(x.unsqueeze(0).unsqueeze(0), scale_factor=1/f, mode=mode, align_corners=align_corners)[0, 0]
```

Here's our "image":

```python
>>> x = torch.tensor([1, 8, 3, 9]).float()
```

To resample by a factor of 2, we try to figure out which values would go between our current values:

```
1   8   3   9
  ?       ?
```

Because it's linear interpolation and we're right between two samples, we use 0.5 * left + 0.5 * right = 0.5 * (left + right) = (left + right) / 2.

Here, (1 + 8) / 2 = 4.5 and (3 + 9) / 2 = 6.

```python
>>> resampled_2 = resample(x, f=2)
>>> resampled_2
tensor([4.5000, 6.0000])
```

If we do that again with our output, we have (4.5 + 6) / 2 = 5.25.

```python
>>> resampled_2_2 = resample(resampled_2, f=2)
>>> resampled_2_2
tensor([5.2500])
```

If we resample directly by a factor of 4, we're in this situation:

```
1   8   3   9
      ?
```

And we get (8 + 3) / 2 = 5.5. Different to the previous 5.25!

```python
>>> resampled_4 = resample(x, f=4)
>>> resampled_4
tensor([5.5000])
```

If we use nearest neighbor interpolation, we only look at the closest sample here.

```
1   8   3   9
  ?       ?
```

As we're right in between, we need to arbitrarily choose which sample we look at. That'd be an implementation detail. Let's see what PyTorch does:

```python
>>> resampled_nn_2 = resample(x, f=2, mode='nearest')
>>> resampled_nn_2
tensor([1., 2.])
```

It seems to use the value from the smaller index. You can start seeing here why nearest neighbour is bad. The output is not a smooth version of the input at all.

If we do this again, we'll get `tensor([1.])` of course.

I was curious to see what would happen with a factor of 4.

```python
>>> resampled_nn_4 = resample(x, f=4, mode='nearest')
>>> resampled_nn_4
```

It just takes the first index! Anyway, let's move on.

Because of the downsampling factor of 2, average pooling is like resampling with linear interpolation:

```python
>>> avg_pooled_2 = avg_pool(x, k=2)
>>> avg_pooled_2
tensor([4.5000, 6.0000])
```

and of course

```python
>>> avg_pooled_2_2 = avg_pool(avg_pooled_2, k=2)
>>> avg_pooled_2_2
tensor([5.2500])
```

Now, average pooling with a kernel of size 4 is different. You need the average of the four values: (1 + 8 + 3 + 9) / 4 = 5.25:

```python
>>> avg_pooled_4 = avg_pool(x, k=4)
>>> avg_pooled_4
tensor([5.2500])
```

From a DSP point of view, there are a few things to take into account.

When you resample a signal with a specific sampling frequency, aliases (copies) of the signal spectrum centered on multiples of the sampling frequency are generated (can be shown with maths 🪄 )

If your sampling frequency is very high, you're good because your aliases are far from your signal. But for downsampling, your frequency is low and the aliases will overlap with your signal, creating artifacts. The solution is to apply a low-pass filter to make your spectrum narrower and then the aliases won't overlap with your signal.

image

To apply a low-pass filter in Fourier space we can multiply the spectrum F(k) of our image f(x) by a step function S(k). That should remove high frequencies, i.e. blur the image (from an old coursework exercise):

image

But a multiplication F(k) × S(k) in Fourier space is mathematically equivalent to a convolution f(x) * s(x) in image space. The inverse Fourier transform is a sinc function.

image

So we could convolve our image with a sinc function for resampling, which is nice but expensive. That's Lanzcos interpolation.

Instead of a fancy sinc function with a wide support, we could use a more boring approximation with a small kernel. How about approximating the sinc with something that looks like a rectangle and convolve that? But wait, that's exactly what we did above!

```
1   8   3   9
_____
  ?       ?


1   8   3   9
        _____
  ?       ?
```

Interestingly, the Fourier transform of a step function is a sinc. If we multiply the spectrum by a sinc, the low-pass filter is not very nice but still does something reasonable. So both 1) resampling with linear interpolation and 2) average pooling are not terrible choices for downsampling! But using large downsampling factors means low sampling frequencies and therefore aliases that will overlap a lot with our image unless we perform aggressive low-pass filtering.

Now, maybe we can use a nicer filter than a sinc in Fourier space. How about a Gaussian? It turns out that the Fourier transform of a Gaussian is a Gaussian, that's why we use a Gaussian kernel for filtering (convolving). The question is how to choose the variance of the Gaussian, but thankfully smart people have written about that, and I cite them in this PR.

---

A bit messy and not super accurate, but hopefully that helps a bit!

fepegar:
Also, this is the test image:

```python
import math

import numpy as np


def zone_plate_3d(
    shape: tuple[int, int, int],
    base_frequency: float = 1,
) -> np.ndarray:
    coords = [
        np.linspace(-1.0, 1.0, n, endpoint=False)
        for n in shape
    ]
    z, y, x = np.meshgrid(*coords, indexing="ij", copy=False)

    r2 = x * x + y * y + z * z
    phase = math.pi * base_frequency * r2
    volume = 0.5 * (1.0 + np.cos(phase))

    return volume

shape =  3 * (256,)
volume = zone_plate_3d(shape, base_frequency=16)
zone_plate = tio.ScalarImage(tensor=volume[None])
zone_plate.plot()
```

image

```python
avg_pool_8 = tio.Lambda(
    lambda x: torch.nn.functional.avg_pool3d(x, kernel_size=8, stride=8)
)

downsampled_zone_plate = avg_pool_8(zone_plate)
downsampled_zone_plate.plot()
```

image


Compare to the image filtered with a Gaussian kernel:

image

Note that the artifacts are less visible when Gaussian smoothing is applied before resampling.

romainVala:
> Note that the artifacts are less visible when Gaussian smoothing is applied before resampling.

of course smothing remove noise, and high spatial frequency (which will change a lot the input image by removing the rings).   it is funny to do the same plot with `base_frequency=4` there are almost no difference (0.03  in average versus 0.22 with 16) 

But my point is that I am not sure this is an aterfact, the avgpool version is the correct low resolution content ...

romainVala:
> I disagree that these methods are "right" or "wrong". I think you just didn't expect resampling with linear interpolation to behave the way it does, and average pooling is closer to what you expected.

may be  but  if can you conclude gaussian smoothing is better then at least  avg pooling is better than linear ressampling (rigth ?)

What is your criteria to say better ?

(if this is to show less artefact, then just smooth more you'll reduce them even more...)


When is say avg pooling is right, it is regarding the following question:

How to downsample, a discrete signal,  so that the resulting low res discrete signal would be the same than the one we would get from sampling the same continuous signal ?

if the input an the output grid are aligned, then the correct downsampling operation should be average pooling

fepegar:
> it is funny to do the same plot with base_frequency=4 there are almost no difference (0.03 in average versus 0.22 with 16)

Yes, that's expected.

Intuitively, as the image is already smooth, the smoothing filter will have little effect on it and downsampling won't make a large difference.

From a theory point of view, the signal spectrum is narrow and therefore the alias spectrum generated at the new sampling frequency is still far away and therefore doesn't overlap with the signal spectrum and no smoothing filter is needed. If we do smooth, the smoothing filter is probably much wider than the signal's spectrum so again nothing changes much.

> may be but if can you conclude gaussian smoothing is better then at least avg pooling is better than linear ressampling (rigth ?)

Generally, yes.

> What is your criteria to say better ? (if this is to show less artefact, then just smooth more you'll reduce them even more...)

Intuitively, using a Gaussian kernel to compute weights for averaging makes more sense (is _better_) than using average pooling (a box kernel): values further away from the new point should contribute less than values that are closer.

From a theory point of view, the Fourier transform of a Gaussian is _better_ than the Fourier transform of a box:

image

The box filter creates zeros and ripples in the Fourier space, which might manifest as artifacts.

fepegar:
> How to downsample, a discrete signal, so that the resulting low res discrete signal would be the same than the one we would get from sampling the same continuous signal ?

This depends on the highest frequency components of the signal. The sampling frequency after downsampling must be at least double the highest frequency of the signal for this to be possible, and downsampling must be done with care. Average pooling would probably not work optimally for this.

> if the input an the output grid are aligned, then the correct downsampling operation should be average pooling

This is not true in the sense that average pooling would be a possible way of doing things, but it's far from optimal.

romainVala:
Yes I agreed to merge, since it is an option, there is no problem

but If I can still use this thread to continue (a bit the discussion)

fepegar:
Thank you, @romainVala

romainVala:
Discussing again this topic with a (physician) collegue, I must admit I am wrong ...  

If I still keep with the objectiv to dowsample the image, in order to obtain the same image that would have been acquired by MRI (at lower resolution)
I need to consider, the sampling process that is perform by MRI, that is acquiring in the fourier domain (k-space)

So this lead to an other alternative, to performe the resampling (respecting MR physic) : 
1_ perform the ifft of the image
2_ keep the center
3_ perform the fft

If I use the first definition of zone_plate_3d
```

factor = 8
shape =  3 * (256,)
volume = zone_plate_3d(shape, base_frequency=16)
image = tio.ScalarImage(tensor=volume[None])
new_shape = shape[0]//factor
center =  shape[0]//2

ifft = torch.fft.ifftshift(torch.fft.ifftn(image.data[0]))
ifft_cut = ifft[(center-new_shape//2):(center+new_shape//2), (center-new_shape//2):(center+new_shape//2), (center-new_shape//2):(center+new_shape//2)]
image_down_fft = torch.fft.fftn(torch.fft.fftshift(ifft_cut)).abs()

tio_img_down_fft = tio.ScalarImage(tensor=image_down_fft[None])
tio_img_down_fft.plot()

```
fft

legend : 
in row order : 1) original image 2) DownSampling (DS) with linear interp 3) DS with avg pooling 4) DS with fft

PR #1360, opened 2025-08-08
Stop copying whole image before cropping (again)

fepegar:
Co-authored-by: Stijn van Wijngaarden <stijn@mriguidance.com>


fepegar:
Thanks for checking!

The memory leak is not an issue since
- https://github.com/TorchIO-project/torchio/pull/1340

What might be nice is adding that check again but without the memory leak. Is that what you mean?

In any case that's indeed out of the scope of this PR.

fepegar:
@allcontributors please add @StijnvWijn for bug

PR #1390, opened 2025-11-24
Fix `_parse_range` with `min` constraint

### LINKED ISSUE #1389: Problem in _parse_range with min_constraint

- [x] I have searched the existing issues

### Bug summary

When setting a `min_constraint` in `Transform._parse_range` and calling with an upper bound only, the upper bound is also used as the lower bound (instead of using the min_constraint there).

The bug is in this line https://github.com/TorchIO-project/torchio/blob/16d3dc669fd13a4db7395c9089885f6b84a30c3a/src/torchio/transforms/transform.py#L334

It should read:
```python
            min_range = -nums_range if min_constraint is None else min_constraint
```

### Code for reproduction

```python
tio.Transform._parse_range(5, 'name', min_constraint=0)
```

### Actual outcome

(5, 5)

### Error messages

```python-traceback

```

### Expected outcome

(0, 5)

### System info

```shell
Platform:   Linux-6.1.0-40-amd64-x86_64-with-glibc2.36
TorchIO:    0.21.0
PyTorch:    2.9.1+cu128
SimpleITK:  2.5.3 (ITK 5.4)
NumPy:      2.3.5
Python:     3.13.8 (main, Oct 10 2025, 12:48:21) [Clang 20.1.4 ]
```

iimog:
Fixes #1389 

**Description**

This pr adds a test that the `min_constraint` is properly used in `_parse_range` and fixes the issue in the line, that caused the bug.

**Checklist**

- Changes are
- [ ] Documentation updated

I did not modify the docstrings or documentation because this change ensures they are accurate. Previously, the described behaviour did not occur (e.g., for RandomNoise, providing a single value for `std` was supposed to apply a random noise level between 0 and `std`, but actually, the maximum value of `std` was always applied). In this way, it could be considered a breaking change if users actively relied on this behavior.

### DIFF
--- a/src/torchio/transforms/augmentation/intensity/random_spike.py
+++ b/src/torchio/transforms/augmentation/intensity/random_spike.py
@@ -41,7 +41,7 @@ class RandomSpike(RandomTransform, IntensityTransform, FourierTransform):
 
     def __init__(
         self,
-        num_spikes: int | tuple[int, int] = 1,
+        num_spikes: int | tuple[int, int] = (1, 1),
         intensity: float | tuple[float, float] = (1, 3),
         **kwargs,
     ):


iimog:
I made the requested change. Is there a way for me to grant you push access to this pr?

PR #1393, opened 2025-12-04
Add RandomBiasFieldDenoise and fix Colin27 exampleFix/colin27 example

Anil970198:
This pull request introduces a new transform, **`RandomBiasFieldDenoise`**, and registers it properly within TorchIO so it can be imported using:

```python
from torchio.transforms import RandomBiasFieldDenoise
````

Additionally, this PR updates the Colin27 dataset usage to avoid a runtime error caused by referencing a non-existent `.subject` attribute.

---

## What's Included

* Added `RandomBiasFieldDenoise` implementation under `augmentation/intensity/`
*  Updated `__init__.py` to expose the new transform via `torchio.transforms`
*  Fixed Colin27 example to correctly access the dataset via indexing
*  Added a small visualization script (`visualize_denoise.py`) to test functionality

---

## Motivation

While exploring TorchIO, the class name `RandomBiasFieldDenoise` appeared referenced in internal modules, but no working implementation or import path existed. Attempting to use it resulted in import errors.

This PR resolves those issues by:

* Adding a functional implementation,
* Ensuring proper registration in import paths,
* Fixing an example affected by the missing import.

This helps prevent confusion for users and improves consistency across augmentation transforms.

---

## Testing

The visualization script was executed using:

```bash
python visualize_denoise.py
```

The script successfully ran and produced before/after image output demonstrating transform behavior.

---

## Additional Notes

This implementation serves as a lightweight placeholder. It may be replaced or extended in the future with a more robust denoising strategy.

### DIFF
--- a/src/torchio/transforms/augmentation/intensity/random_biasfield_denoise.py
+++ b/src/torchio/transforms/augmentation/intensity/random_biasfield_denoise.py
@@ -1,10 +1,7 @@
 from torchio.data.subject import Subject
 
-<<<<<<< HEAD
-=======
 from ...transform import Transform
 
->>>>>>> e8e8035 (Apply Ruff autofix and sort imports)
 
 class RandomBiasFieldDenoise(Transform):
     """


fepegar:
@Anil970198, can you explain these changes? Were these commits created by a bot?

PR #1463, opened 2026-05-29
Fix comment in `nib_to_sitk`

vcasellesb:
**Description**

I have updated a comment which I believe was confusing and misleading. On `torchio/data/io.py`, function `nib_to_sitk`, the shape that is provided in the comment next to `array.transpose()` does not agree with the shapes that are provided as possibilities at the beginning of the function.

**Checklist**

- [ ] I have read the [`CONTRIBUTING`](https://github.com/TorchIO-project/torchio/blob/main/CONTRIBUTING.md) docs and have a developer setup ready
- Changes are
- [ ] Documentation updated

### DIFF
--- a/src/torchio/data/io.py
+++ b/src/torchio/data/io.py
@@ -318,7 +318,7 @@ def nib_to_sitk(
         array = array[..., 0]
     if not is_multichannel and not force_4d:
         array = array[0]
-    array = array.transpose()  # (W, H, D, C) or (W, H, D)
+    array = array.transpose()  # (D, H, W, C) or (H, W, C) or (D, H, W) or (H, W)
     image = sitk.GetImageFromArray(array, isVector=is_multichannel)
 
     origin, spacing, direction = get_sitk_metadata_from_ras_affine(


vcasellesb:
> ## Pull request overview
> Corrects an inline comment in `nib_to_sitk` that described only one possible output shape after `array.transpose()`, when in fact four shape variants are possible depending on dimensionality and channel count. The new comment enumerates all four cases consistent with the function's documented input possibilities.
> 
> **Changes:**
> 
> * Updates the shape annotation comment after `array.transpose()` to list all four valid post-transpose shapes: `(D, H, W, C)`, `(H, W, C)`, `(D, H, W)`, `(H, W)`.
> 
> 💡 [Add Copilot custom instructions](/TorchIO-project/torchio/new/main?filename=.github/instructions/*.instructions.md) for smarter, more guided reviews. [Learn how to get started](https://docs.github.com/en/copilot/customizing-copilot/adding-repository-custom-instructions-for-github-copilot).

This is wrong! The error is not in the number of possible output shapes, but in the numpy shapes being wrong themselves (besides not accounting for all possibilities). The comment does not transpose all dimensions as the homonymous method does.

fepegar:
Thanks for the fix, @vcasellesb. I agree the clanker is wrong.

fepegar:
@allcontributors please add @vcasellesb for doc

PR #1511, opened 2026-09-12
Respect image selection in label transforms

### LINKED ISSUE #1508: Label transforms ignore include/exclude filters
### Description

`RemapLabels`, `RemoveLabels`, and `OneHot` modify label maps that should be excluded by the `include` or `exclude` arguments.

### Reproduction

```python
import torch
import torchio as tio

subject = tio.Subject(
    a=tio.LabelMap(torch.ones(1, 2, 2, 2)),
    b=tio.LabelMap(torch.ones(1, 2, 2, 2)),
)

result = tio.RemapLabels({1: 9}, include=["a"])(subject)

print(result.a.data.unique())  # tensor([9.])
print(result.b.data.unique())  # tensor([9.]), expected tensor([1.])
```

The same problem occurs with:

```python
tio.RemapLabels({1: 9}, exclude=["b"])
tio.RemoveLabels([1], include=["a"])
tio.RemoveLabels([1], exclude=["b"])
tio.OneHot(include=["a"])
tio.OneHot(exclude=["b"])
```

### Expected behavior

Only the selected label maps should change.
Excluded label maps should retain their original values and shape.

### Possible cause

These transforms iterate over `batch.images.items()` directly instead of using `self._get_images(batch)`, which applies the include/exclude filters.

### Version

Reproduced on `main`, commit `2b019d2a9659c838408c73ae77327ac5eb87b190`, TorchIO `2.0.0a2`, using synthetic CPU tensors.

Shy7777:
Fixes #1508.

**Description**

With two label maps, `RemapLabels({1: 9}, include=["a"])` currently changes `b` as well. `RemoveLabels` and `OneHot` have the same problem. These transforms now use the existing image-selection helper so that excluded maps keep their values and shape.

This also applies to the recorded inverse of `OneHot`: an excluded map that already has multiple channels should retain those channels when another map is decoded. The regression tests cover include/exclude combinations, empty selections, individual subjects, batches, recorded inverses, and preservation of the original input.

Validation on macOS with Python 3.13: **1476 passed, 4 skipped**, including all 44 new regression cases. The skips are two CUDA tests and two optional ipyniivue tests. Ruff checks pass. All applicable pre-commit hooks pass except type checking, which reports an existing `ndarray.max` type error in `transforms/intensity/blur.py:276`; the same error reproduces on the unchanged base commit.

**Checklist**

- [x] I have read the [`CONTRIBUTING`](https://github.com/TorchIO-project/torchio/blob/main/CONTRIBUTING.md) docs and have a developer setup ready
- Changes are
- [ ] Documentation updated

The existing include/exclude documentation already describes the intended behavior, so no documentation changes are needed.


fepegar (approved):
LGTM. Thank you, @Shy7777!

fepegar:
@allcontributors please add @Shy7777 for bug.

fepegar:
@allcontributors please add @Shy7777 for bug.

